PR #32, opened 2020-01-02
Ot samda 1

olliethomas:


thomkeh:
Can I add a commit to this?

olliethomas:
of course. No rush at my end

thomkeh:
You had use `Float` instead of `_Float` and that made mypy just accept any input. It's a bit concerning that the tests didn't catch this. I opened #33 to discuss better tests that would catch such an error.

PR #40, opened 2020-01-20
expand use cases and add linalg module

olliethomas:


olliethomas:
I'm not 100% on all of these, need to try them out on other packages (eg EthicML)

numpy-stubs/__init__.pyi
```diff
@@ -660,7 +673,10 @@ def unique(
 def unique(
     a: ndarray[_DType], return_inverse: bool = ..., axis: Optional[int] = ...
 ) -> Tuple[ndarray[_DType], ndarray[_DType]]: ...
+@overload
 def where(condition: ndarray[bool_], x: ndarray[_DType], y: ndarray[_DType]) -> ndarray[_DType]: ...
+@overload
+def where(condition: ndarray[bool_]) -> ndarray[_DType]: ...
```
thomkeh:
```suggestion
def where(condition: ndarray[bool_]) -> ndarray: ...
```

It's pointless to have an unbounded TypeVar.

numpy-stubs/linalg.py
```diff
@@ -0,0 +1,7 @@
+from typing import Tuple
+
+from numpy import ndarray, _Int, float64, _DType
+
+
+def slogdet(x: ndarray[_DType]) -> Tuple[_Int, ndarray[float64]]:
```
thomkeh:
```suggestion
def slogdet(x: ndarray[_DType]) -> Tuple[int, ndarray[float64]]:
```

same here. is `int` the right type? or is it `np.integer`.

pandas-stubs/core/indexes.pyi
```diff
@@ -16,14 +16,21 @@ class Index(Generic[_T]):
     @overload
     def __getitem__(self, idx: int) -> _T: ...
     @overload
+    def __getitem__(self, idx: Index[_T]) -> Index[_T]: ...
+    @overload
+    def __getitem__(self, idx: Series[bool]) -> _T: ...
```
thomkeh: meh. you should use `.loc` instead for `Series[bool]`, but okay...

numpy-stubs/__init__.pyi
```diff
@@ -205,11 +205,11 @@ class ndarray(Generic[_DType]):
     ) -> None: ...
     # for some reason, you can not use _Float to narrow down the type of ndarray here:
     @overload
-    def tolist(self: Union[ndarray[float32], ndarray[float64]]) -> Sequence[float]: ...
+    def tolist(
+        self: Union[ndarray[bool_], ndarray[int8], ndarray[int16], ndarray[int32], ndarray[int64]]
```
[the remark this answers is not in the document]
olliethomas: bool is in with the ints as numpy treats bool as a (small) int

numpy-stubs/linalg.py
olliethomas:
```
a, b = np.linalg.slogdet(np.eye(500) * 0.1)
type(a) # <class 'numpy.float64'>
```
Looks like it actually returns np.float64

pandas-stubs/core/indexes.pyi
olliethomas:
```
error: "Index[str]" has no attribute "iloc"
error: "Index[str]" has no attribute "loc"
```

olliethomas: This is the line for context
`list(data.columns.iloc[(data.apply(np.std) <= threshold)])`

thomkeh: Oh I see. sorry

PR #48, opened 2020-01-29
added figure and subplot_adjust

Bradley-Butcher:
Trying to stub plt.figure and plt.subplots_adjust.

Uncertain about line 77:
`    **kwargs: Optional[Any]) -> Figure:
`

### DIFF
--- a/matplotlib-stubs/pyplot.pyi
+++ b/matplotlib-stubs/pyplot.pyi
@@ -74,7 +74,8 @@ def figure(
     frameon: Optional[bool] = ...,
     FigureClass: Type[Figure] = ...,
     clear: Optional[bool] = ...,
-    **kwargs: Optional[Any]) -> Figure:
+    **kwargs: Optional[Any],
+) -> Figure: ...
 @overload
 def subplots_adjust(
     left: Optional[float] = ...,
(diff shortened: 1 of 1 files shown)


Bradley-Butcher:
Hold up didn't realize you need multiple overload definitions

matplotlib-stubs/pyplot.pyi
```diff
@@ -64,6 +64,25 @@ def subplots(
     dpi: int = ...,
     figsize: Tuple[float, float] = ...,
 ) -> Tuple[Figure, List[List[Axes]]]: ...
+def figure(
+    num: Optional[Union[int, str]] = ...,
+    figsize: Optional[List[float]] = ...,
```
thomkeh: Maybe do it like this instead?

```suggestion
    figsize: Optional[Tuple[float, float]] = ...,
```

But I don't really mind either way.

matplotlib-stubs/pyplot.pyi
```diff
@@ -64,6 +64,25 @@ def subplots(
...
+    dpi: Optional[int] = ...,
+    facecolor: Optional[str] = ...,
+    edgecolor: Optional[str] = ...,
+    frameon: Optional[bool] = ...,
```
thomkeh: I think looking at the documentation, that this cannot be `None`.

```suggestion
    frameon: bool = ...,
```

matplotlib-stubs/pyplot.pyi
```diff
@@ -64,6 +64,25 @@ def subplots(
...
+    FigureClass: Type[Figure] = ...,
+    clear: Optional[bool] = ...,
```
thomkeh: Same here

```suggestion
    clear: bool = ...,
```

thomkeh:
I'll just do the small changes and merge it.

thomkeh:
The documentation said that `frameon` and `clear` are "optional", but they mean it in the sense that these parameters are not required; and not in the sense that these parameters can be `None`.

PR #57, opened 2020-02-11
Additional type signatures for DataFrame and Series methods

aguillon:
Hi,

I'm not very sure about the signatures in Series and would be happy to discuss them. I'm still inexperienced with Mypy.


aguillon:
Woops, my bad. By the way, how do I run tests on my github fork?

thomkeh:
Sorry for the late reply.

We now have some instructions here: https://github.com/predictive-analytics-lab/data-science-types#checking-compliance-with-mypy

Basically, you have to install mypy and then just do `mypy tests`. It's not really a proper unit test. It's just some python modules that are type checked with mypy. And if mypy doesn't complain about them, then the test passes.

pandas-stubs/core/series.pyi
```diff
@@ -72,6 +72,26 @@ class Series(Generic[_DType]):
...
+        self, to_append: Tuple[Series], ignore_index: bool = ..., verify_integrity: bool = ...
+    ) -> Series: ...
+    @overload
+    def apply(
+        self, func: Callable, convert_dtype: bool = ..., args: Tuple = ..., **kwargs: Any
+    ) -> Series: ...
+    @overload
+    def apply(
+        self, func: Callable, convert_dtype: bool = ..., args: Tuple = ..., **kwargs: Any
+    ) -> DataFrame: ...
```
thomkeh: The problem is here probably. you have two definitions of `apply()` that are completely identical, but they return different types. That's not allowed for overloading (because then you cannot know what the return type is).

pandas-stubs/core/series.pyi
```diff
@@ -72,6 +72,26 @@ class Series(Generic[_DType]):
...
+        self, to_append: Series, ignore_index: bool = ..., verify_integrity: bool = ...
+    ) -> Series: ...
+    @overload
+    def append(
+        self, to_append: List[Series], ignore_index: bool = ..., verify_integrity: bool = ...
+    ) -> Series: ...
+    @overload
+    def append(
+        self, to_append: Tuple[Series], ignore_index: bool = ..., verify_integrity: bool = ...
+    ) -> Series: ...
```
thomkeh: This code is technically correct, but you can make it shorter with `Union`:

```python
def append(
        self, to_append: Union[Series, List[Series], Tuple[Series, ...]], ignore_index: bool = ..., verify_integrity: bool = ...
    ) -> Series: ...
```

You only need to use `overload` when the return type differs for different arguments.

So, here you need overload:

```python
@overload
def f(x: int) -> int: ...
@overload
def f(x: str) -> str: ...
```

But here you don't:
```python
@overload
def g(x: int) -> float: ...
@overload
def g(x: str) -> float: ...
```
you can do this instead:
```python
def g(x: Union[int, str]) -> float: ...
```

---

Also, you need to use `Tuple[Series, ...]` instead of `Tuple[Series]`. `Tuple[Series]` means a tuple with only **one** element. `Tuple[Series, ...]` means a tuple with an unknown number of elements.

You can also just use `Sequence[Series]`. `Sequence[T]` is basically equivalent to `Union[List[T], Tuple[T, ...]]`.

aguillon:
Thanks a lot for your explanations. I updated the commits and ran the tests. I think these signatures are correct; it's only signatures I need, which explains the fact that I add them one after another with very small commits. I hope it's ok with you?

I have another question, if I may: is there a reason why so many signatures are specific about the type of the containee, e.g. `Series[bool]` and such? Is this a "good enough" solution instead of making the types too complicated with type variables?

thomkeh:
Code looks good! I'll squash all the commits anyway.

>  I have another question, if I may: is there a reason why so many signatures are specific about the type of the containee, e.g. Series[bool] and such? Is this a "good enough" solution instead of making the types too complicated with type variables?

Do you mean like in
```python
def isnull(self) -> Series[bool]: ...
```
?

There the result really will be a series object that contains bools.

Other times we have something like this:
```python
@overload
def repeat(x: int, repeats: int) -> np.ndarray[np.int64]: ...
@overload
def repeat(x: float, repeats: int) -> np.ndarray[np.float64]: ...
```
Here the problem is that we want to say: if you pass `repeat()` an `int`, then you get an array of `np.int64`; whereas for `float` you get `np.float64`. And the only way that I found how to express this, is by using these overloads.

thomkeh:
Should I merge this PR? Or do you want to add more commits? Both are fine for me.

aguillon:
> > I have another question, if I may: is there a reason why so many signatures are specific about the type of the containee, e.g. Series[bool] and such? Is this a "good enough" solution instead of making the types too complicated with type variables?
> 
> Do you mean like in
> 
> ```python
> def isnull(self) -> Series[bool]: ...
> ```

I was thinking more about signatures like:
```python
@overload
def astype(self, dtype: Type[int]) -> Series[int]: ...
@overload
def astype(self, dtype: Type[float]) -> Series[float]: ...
```

Is there a reason why it's not written as
```python
def astype(self, dtype: Type[_DType]) -> Series[_DType]: ...
```
where `_DType` is a type variable?

On other places, the signatures just return `Series`, which is a synonym for `Series[Any]`. Shouldn't this be `Series[_DType]`?

> Should I merge this PR? Or do you want to add more commits? Both are fine for me.

I think I won't make any new PR until next week, so you might as well merge if you prefer.

Thank you for your time!

thomkeh:
I see. I also cannot see a reason why `astype` was typed this way. It might just be a mistake. Or there is some weird problem that appears when you use `_DType`. That sometimes happens as well. I will try it out with `_DType` and see.

In general I'm not completely happy how we treat the dtype of Series, because if you set the dtype to int, then:

```python
In [2]: pd.Series([0, 1, 2], dtype=int)
Out[2]:
0    0
1    1
2    2
dtype: int64
```

the actual dtype is `int64`.

Whereas
```python
In [6]: pd.Series([0, 1, 2], dtype="int32")
Out[6]:
0    0
1    1
2    2
dtype: int32
```

For numpy, I did it properly and have all the dtypes (`np.int16`, `np.int32`, ...), but it was a lot of work, so I'm not keen on doing it again. But it should be done at some point.

Pandas 1.0 actually [introduced new dtypes](https://pandas.pydata.org/pandas-docs/version/1.0.0/whatsnew/v1.0.0.html): `StringDtype`, `BooleanDtype`, `Int64Dtype`, `Int32Dtype`. So maybe we should just use those. Although they weirdly show up as type `object` for now:

```python
In [3]: pd.Series([0, 1, 2], dtype=pd.Int64Dtype)
Out[3]:
0    0
1    1
2    2
dtype: object

In [4]: pd.Series([0, 1, 2], dtype=pd.Int64Dtype).dtype
Out[4]: dtype('O')
```

PR #58, opened 2020-02-12
overload np getitem and pd where

SaraR-1:


numpy-stubs/__init__.pyi
```diff
@@ -678,6 +678,8 @@ def unique(
 @overload
 def where(condition: ndarray[bool_], x: ndarray[_DType], y: ndarray[_DType]) -> ndarray[_DType]: ...
 @overload
+def where(condition: ndarray[bool_], x: Optional[int], y: Optional[int]) -> ndarray[_DType]: ...
```
thomkeh: Do you really need `Optional` here? Is it useful to pass `None` to this function?

I think we can probably do this:

```suggestion
def where(condition: ndarray[bool_], x: int, y: int) -> ndarray[int64]: ...
```

PR #60, opened 2020-02-13
Added usecols parameter to read_csv and added read_sql method stub.

ZHSimon:
Sorry if I'm doing this wrong - first time trying to actually participate in a project like this!


pandas-stubs/core/series.pyi
```diff
@@ -107,7 +107,7 @@ class Series(Generic[_DType]):
     def std(self) -> float: ...
     def sum(self) -> float: ...
     def to_dict(self) -> Dict[Union[int, _str], _DType]: ...
-    def to_frame(self, name: Optional[_str]) -> DataFrame: ...
+    def to_frame(self, name: Optional[_str] = None) -> DataFrame: ...
```
thomkeh: Default values should not appear in type stubs. Use `...` instead.

```suggestion
    def to_frame(self, name: Optional[_str] = ...) -> DataFrame: ...
```

pandas-stubs/core/frame.pyi
```diff
@@ -103,12 +105,14 @@ class DataFrame:
     def drop(self, index: Union[List[_str], Index], axis: _AxisType = ...) -> DataFrame: ...
     def drop_duplicates(self, keep: Union[_str, bool] = ...) -> DataFrame: ...
     def dropna(self, axis: int = ..., how: _str = ...) -> DataFrame: ...
+    def fillna(self, value: Union[Scalar, Dict, Series, DataFrame], method: str = None, axis: Union[str, int] = None, inplace: bool = False, limit: int = None, downcast: Dict = None) -> Union[DataFrame, None]: ...
```
thomkeh: Same here; replace the default values by `...`.

There is another problem here: you are _returning_ a `Union`. This should be avoided if possible. And in this case it is possible to avoid it, with `@overload`:
```suggestion
    @overload
    def fillna(self, value: Union[Scalar, Dict, Series, DataFrame], method: str = ..., axis: Union[str, int] = ..., inplace: Literal[True], limit: int = ..., downcast: Dict = ...) -> None: ...
    @overload
    def fillna(self, value: Union[Scalar, Dict, Series, DataFrame], method: str = ..., axis: Union[str, int] = ..., inplace: Literal[False] = ..., limit: int = ..., downcast: Dict = ...) -> DataFrame: ...
```

The way it works is this: if you do not specify `inplace` or set it to `False` (`inplace=False`), then the second overload is used and a DataFrame is returned. If instead you set `inplace=True`, then the first overload is used and nothing is returned.

pandas-stubs/core/frame.pyi
```diff
@@ -13,8 +13,10 @@ from typing import (
     Iterator,
     Sequence,
     Generator,
+    Iterable,
+    Series,
 )
-from typing_extensions import Literal
+from typing_extensions import Literal, Scalar
```
thomkeh: What is `Scalar`? I cannot find anything about it. If you wanted to have `Union[int, float]`, then just use `float`; it counts for both.

pandas-stubs/_core.pyi
```diff
@@ -36,6 +36,17 @@ def read_csv(
     delim_whitespace: bool = ...,
     names: Optional[List[str]] = ...,
     dtype: Optional[Any] = ...,
+    usecols: Optional[Union[List], Callable] = ...,
+) -> DataFrame: ...
+def read_sql(
+    sql: Union[str, Any],
+    con: Union[Any, str] = ...,
+    index_col: Optional[Union[str, List[str]]] = ...,
+    coerce_float: bool = True,
```
thomkeh:
```suggestion
    coerce_float: bool = ...,
```

ZHSimon:
> Thanks a lot for submitting a PR! Definitely appreciated.
> 
> I left some comments on the code. Some style issues and some details on `@overload`.
> 
> The other bigger problem is that you submitted the PR not targeting the master branch but targeting the branch `releases/0.2.4`. Can you change that somehow? Maybe you have to re-submit the PR.

Looks like I will need to resubmit the branch. I'll address these and resubmit shortly!

pandas-stubs/core/frame.pyi
ZHSimon: I pulled that term directly from the documentation - ["value - scalar, dict, Series, or DataFrame
Value to use to fill holes (e.g. 0), alternately a dict/Series/DataFrame of values specifying which value to use for each index (for a Series) or column (for a DataFrame). Values not in the dict/Series/DataFrame will not be filled. This value cannot be a list."](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.fillna.html?highlight=dataframe%20fillna#pandas.DataFrame.fillna)

TBH, I'm not sure if just allowing a float would work.

PR #61, opened 2020-02-13
Small additions to Pandas

ZHSimon:
Added read_sql method, added usecols argument; added fillna and isin methods to DataFrame; added parameter information to to_frame in Series.


olliethomas:
Thanks for the PR - looks like it's failing on black. Should be an easy fix, check out this section of the readme for guidance (and if it can be improved on, please let us know) https://github.com/predictive-analytics-lab/data-science-types#formatting-with-black

ZHSimon:
> Thanks for the PR - looks like it's failing on black. Should be an easy fix, check out this section of the readme for guidance (and if it can be improved on, please let us know) https://github.com/predictive-analytics-lab/data-science-types#formatting-with-black

Looks like that wasn't the only issue! Sorry, this looks like a perfect time to do a squash merge?

olliethomas:
There's something I'm not entirely happy with, but don't know the exact way forward. Let me explain,
```
sql: Union[str, Any],
con: Union[Any, str] = ...,
```
The `Union` type is unnecessary here, but I can understand why you have it. `Any` is a special type that is basically a `Union`  of every type possible, so `Any` already includes `str`, but also includes `float`, `int`, `list` etc. making it too permissive to be useful. 

Looking at the pandas docs I can see that the introduction of `Any` is probably to cover the case of a SQLAlchemy Selectable type being passed as the `sql` arg and a SQLAlchemy connectable type being passed as the `con` arg.

The reason I say that I don't know exactly how to proceed right now is that [Dropbox actually provide type stubs for SQLAlchemy](https://github.com/dropbox/sqlalchemy-stubs). However, the installation looks like it involves a mypy plugin and I don't know much about them, so without experimenting, I don't know if we want to either add this as a dependency, or handle this in some other way.

ZHSimon:
This decision seems to be outside my scope as a contributor.

I did figure out where to import those two classes from - sqlalchemy.connectors contains Connector, and sqlalchemy.sql contains Selectable. I am not sure, however, if that's better or worse than using their stubs package and plugin.

thomkeh:
> This decision seems to be outside my scope as a contributor.

Yes, sorry!

I'll merge this now. Thanks again!

olliethomas:
Agreed. Let’s get this merged in and we can figure out how to best handle the SQLAlchemy later. Thank you for the contribution.

PR #62, opened 2020-02-14
Additional small additions to DataFrame, and one for Series, in Pandas stub

ZHSimon:
I've made a few more additions to the stub, fleshing it out as I found I needed more for my work.

As a note, the stub for the __getitem__ method in Series is incorrect - I tried to find a way to fix it, but was unable to create a cover-all stub for it, only one that solved my specific issue.  I don't know how I'd stub out a return type of Series[index] without knowing what the Series is containing, so I'm not sure of any elegant solutions to this.


ZHSimon:
Turns out this fix isn't quite ready for my work just yet, so I'll resubmit once I've figured it out.

pandas-stubs/core/frame.pyi
```diff
@@ -112,7 +117,17 @@ class DataFrame:
         inplace: bool = ...,
```
thomkeh: This probably needs to change to `inplace: Literal[False] = ...`.

pandas-stubs/core/series.pyi
```diff
@@ -59,7 +59,7 @@ class Series(Generic[_DType]):
     @overload
     def __getitem__(self, idx: Union[List[_str], Index[int], Series, slice, str]) -> Series: ...
     @overload
-    def __getitem__(self, idx: int) -> [float]: ...
```
ZHSimon: It turns out these accidental brackets were what led me to believe my issue was solved. I've got a Series of Series, and I'm getting to a nested series of it via its numerical index. I'd change the return type from float to Series, but that would just break it for whoever set it to float in the first place. The correct solution would be to find a way to stub it so it returns the dtype of that Series, and I'm not sure how to do that.

thomkeh: Right, you should just be able to do what you say:

```python
def __getitem__(self, idx: int) -> _DType: ...
```

That's what the dtypes are for. See for example how we do it for numpy: https://github.com/predictive-analytics-lab/data-science-types/blob/master/numpy-stubs/__init__.pyi#L263

ZHSimon: Thank you! I'll work that out now.

PR #63, opened 2020-02-14
Small additions to DataFrame and Series

ZHSimon:
I've made a few more additions to the stub, fleshing it out as I found I needed more for my work. I've corrected the issue I found - thanks again, thomkeh! - and hope that others can benefit from this work.

Thank you!


pandas-stubs/core/frame.pyi
```diff
@@ -104,15 +108,26 @@ class DataFrame:
...
+        method: _str = ...,
+        axis: Union[_str, int] = ...,
+        inplace: Literal[False] = ...,
+        limit: int = ...,
+        downcast: Dict = ...,
+    ) -> DataFrame: ...
+    @overload
     def fillna(
         self,
+        inplace: Literal[True],
```
thomkeh: Did you need to move this line to make it work? It destroys the symmetry a bit :)

pandas-stubs/core/frame.pyi
```diff
@@ -48,7 +48,10 @@ class DataFrame:
     def __getitem__(self, idx: _str) -> Series: ...
     @overload
     def __getitem__(
-        self, idx: Union[Series, Tuple[_str, ...], List[_str], Index[_str], _np.ndarray[_np.str_]]
+        self,
+        idx: Union[
+            Series, Tuple[_str, ...], List[_str], Index[_str], _np.ndarray[_np.str_], DataFrame
```
thomkeh: Do you have an example for when you need to pass a `DataFrame` here? Does the `DataFrame` contain strings?

thomkeh:
(Btw, my name is Thomas and my family name starts with "Keh", that's why my Github name is thomkeh.)

pandas-stubs/core/frame.pyi
ZHSimon: Agreed, it looks weird; however, moving it to the right place throws an error: 'non-default argument follows default argument'

pandas-stubs/core/frame.pyi
ZHSimon: Unfortunately, I can't put the code I'm working on up on github, but I needed to add that as an option. The data frames I'm working with are... nested in weird ways that I don't quite understand yet. I'll have to bug the rest of the team about why they're doing things this way.

ZHSimon: `frame = frame[frame['Framename'].to_frame().isin(setOfFrames)]`

Like I said, weird.

pandas-stubs/core/frame.pyi
thomkeh: oh, of course. yeah, makes sense.

pandas-stubs/core/frame.pyi
thomkeh: I see... I'm guessing `frame['Framename'].to_frame().isin(setOfFrames)` returns some kind of mask? so, a dataframe where each element is either True or False?

If it is a mask, then you should use `.loc` instead [as the documentation recommends](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.loc.html). This is a typical example from the documentation:

```python
df.loc[df['shield'] > 6]
```

I want to claim that we support this already, but looking at the code, we actually don't support it. Let me add a PR for that.

I will think about this a bit more and reply to you tomorrow. The other changes in this PR can definitely be merged.

ZHSimon: I'll be able to ask my coworkers in person about this on Monday, so I'll have a more solid grasp of why things are the way they are after that.

ZHSimon: After rereading the code - and I still need to confirm this with the team - I believe this line culls out all the values which aren't also in the setOfFrames.

ZHSimon: I reverted the addition and used .loc in its place.

thomkeh:
If you want, I can merge this now.

ZHSimon:
Please do!

ZHSimon:
Quick question - do you have any idea when the Master branch will be released? I know we can get it directly via pip install git+etc, just wondering when it'll be pushed out.

thomkeh:
Usually we just do a release whenever we need it for something. Maybe we should introduce a more regular release schedule. I'll do a release for you today.

ZHSimon:
> I'll do a release for you today.

Thank you very much!

thomkeh:
https://pypi.org/project/data-science-types/0.2.5/

ZHSimon:
Thank you! I've updated my setup.py accordingly, and now my team can simply
pull down the code and use it just fine. Thanks again!

PR #64, opened 2020-02-14
Improve how .loc and .iloc work

thomkeh:


thomkeh:
@ZHSimon your example `frame = frame[frame['Framename'].to_frame().isin(setOfFrames)]` works now. I thought it could be done with `.loc` but it cannot.

I ended up rewriting a lot of the indexing code. It should be much more correct now.

@olliethomas I might have broken stuff in your code. I'll have a look on Monday.

PR #68, opened 2020-02-18
Ot post 2.5

olliethomas:
updated pandas indexing and numpy stubs


pandas-stubs/core/indexing.pyi
```diff
@@ -56,8 +56,14 @@ class _LocIndexerFrame:
     @overload
     def __getitem__(self, idx: _MaskType,) -> DataFrame: ...
     @overload
+    def __getitem__(self, idx: List[int],) -> DataFrame: ...
```
thomkeh: is this for dataframes with numeric indices? can you not use `.iloc` for this?

olliethomas: According to the tests, you're absolutely right

pandas-stubs/core/indexing.pyi
```diff
@@ -58,6 +58,10 @@ class _LocIndexerFrame:
     @overload
     def __getitem__(self, idx: _StrLike,) -> Series: ...
     @overload
+    def __getitem__(self, idx: List[_StrLike],) -> Series: ...
```
thomkeh: I just tried it out and this returns a dataframe
```suggestion
    def __getitem__(self, idx: List[_StrLike],) -> DataFrame: ...
```

pandas-stubs/core/indexing.pyi
```diff
@@ -58,6 +58,10 @@ class _LocIndexerFrame:
...
+    @overload
+    def __getitem__(self, idx: Tuple[_MaskType, Union[_MaskType, List[str]]]) -> DataFrame: ...
```
thomkeh: isn't this one capture by

```python
def __getitem__(
        self, idx: Tuple[Union[_MaskType, List[str]], Union[_MaskType, List[str]]],
    ) -> DataFrame: ...
```
?

(see below)

pandas-stubs/core/indexing.pyi
```diff
@@ -69,6 +73,14 @@ class _LocIndexerFrame:
     @overload
     def __setitem__(self, idx: _StrLike, value: Union[float, Series, _np.ndarray],) -> None: ...
     @overload
+    def __setitem__(
+        self, idx: Tuple[_np.int64, str], value: Union[_np.ndarray, List[int], float]
```
thomkeh: if you're indexing with an `int` (or rather `_np.int64`), then you should probably be using `.iloc`.

(And why only allow `List[int]` as value? what about `List[float]`?)

thomkeh:
indexing in pandas is just such a mess... maybe I should give up and allow all kinds of indexing for any of the indexing methods...

thomkeh:
I rebased this branch. you will have to do:

```
git checkout ot-post-2.5
git fetch
git reset --hard origin/ot-post-2.5
```

pandas-stubs/core/indexing.pyi
olliethomas: Yes, you're right

The reason it was added was because
```
@overload
def __getitem__(self, idx: Tuple[_StrLike, _StrLike],) -> float: ...
```

was capturing the case 
```
def __getitem__(self, idx: Tuple[_MaskType, Union[_MaskType, List[str]]]) -> DataFrame: ...
```

If the order of the overloads is changed, so that the order is now
```
@overload
def __getitem__(
    self, idx: Tuple[Union[_MaskType, List[str]], Union[_MaskType, List[str]]],
) -> DataFrame: ...
@overload
def __getitem__(self, idx: Tuple[_StrLike, _StrLike],) -> float: ...
```

then the underlying problem is dealt with

pandas-stubs/core/indexing.pyi
olliethomas: {ValueError}Location based indexing can only have [integer, integer slice (START point is INCLUDED, END point is EXCLUDED), listlike of integers, boolean array] types

pandas-stubs/core/indexing.pyi
```diff
@@ -58,17 +58,21 @@ class _LocIndexerFrame:
...
     ) -> DataFrame: ...
+    @overload
+    def __getitem__(self, idx: Tuple[_StrLike, _StrLike],) -> float: ...
     # set item
     @overload
     def __setitem__(self, idx: _MaskType, value: Union[float, _np.ndarray, DataFrame],) -> None: ...
     @overload
     def __setitem__(self, idx: _StrLike, value: Union[float, Series, _np.ndarray],) -> None: ...
     @overload
+    def __setitem__(self, idx: Tuple[_IndexType, str], value: Union[_IndexType, float]) -> None: ...
```
olliethomas: I don't really like that this `float` is here

tests/pandas_test.py
```diff
@@ -32,8 +32,13 @@ def test_frame_loc() -> None:
     b: pd.DataFrame = df.loc[s > 6]
     c: float = df.loc["cobra", "shield"]
     d: pd.DataFrame = df.loc[df["shield"] > 6, ["max_speed"]]
+    e: pd.DataFrame = df.loc[["cobra", "viper"]]
+    assert_type(e, pd.DataFrame)
     df.loc[["viper", "sidewinder"], ["shield"]] = 50.0
     df.loc["cobra"] = 10.0
+    df.loc[["cobra", "viper"]] = df
+
+    df.loc[[0, 1], "shield"] = 4.0
```
thomkeh: This throws an error at runtime, but is allowed by the types.

I think the problem is that the index of this dataframe here is based on strings, but in your case the index is integers.

pandas-stubs/core/indexing.pyi
thomkeh: I think it's more correct than the `_IndexType` to be honest.

thomkeh:
Maybe trying to add types for pandas was a bad idea from the start. This whole indexing just really sucks.

tests/pandas_test.py
```diff
@@ -17,6 +17,9 @@ def assert_type(v: T, t: Type[T]) -> None:
     index=["cobra", "viper", "sidewinder"],
     columns=["max_speed", "shield"],
 )
+fd: pd.DataFrame = pd.DataFrame(
+    [[1.0, 2.0], [4.0, 5.0], [7.0, 8.0]], columns=["max_speed", "shield"],
+)
```
olliethomas: added `fd` to reflect dataframe where `loc[Tuple, str]` is being called

tests/pandas_test.py
```diff
@@ -32,8 +35,13 @@ def test_frame_loc() -> None:
     b: pd.DataFrame = df.loc[s > 6]
     c: float = df.loc["cobra", "shield"]
     d: pd.DataFrame = df.loc[df["shield"] > 6, ["max_speed"]]
+    e: pd.DataFrame = df.loc[["cobra", "viper"]]
+    assert_type(e, pd.DataFrame)
     df.loc[["viper", "sidewinder"], ["shield"]] = 50.0
     df.loc["cobra"] = 10.0
+    df.loc[["cobra", "viper"]] = df
+
+    fd.loc[[0, 1], "shield"] = 4.0
```
olliethomas: using `fd` instead of `df`

tests/pandas_test.py
olliethomas: updated in tests

thomkeh:
Great! Sorry for being to nitpicky.

PR #70, opened 2020-02-19
Minor Additions to Series

ZHSimon:
It turns out that I missed two methods in the code I was working on, and need to add them - so here I am with my code in hand, asking for another PR!

This adds the to_csv() method and __or__ method stubs to Series.pyi.


ZHSimon:
For the record, I'm not requesting another release - you were kind enough to do one for me already, and I've found a workaround that will last until your next release, whenever that may be. Thanks!

pandas-stubs/core/series.pyi
```diff
@@ -64,6 +65,7 @@ class Series(Generic[_DType]):
     def __iter__(self) -> Iterator[_DType]: ...
     def __len__(self) -> int: ...
     def __truediv__(self, other: object) -> Series: ...
+    def __or__(self, other: Series) -> Series: ...
```
olliethomas:
```suggestion
    def __or__(self, other: Series) -> Series[bool]: ...
```

olliethomas: @thomkeh

PR #72, opened 2020-02-25
Extend the type hint for pandas.DataFrame.dropna()

rpgoldman:
Add support for the `subset` argument, and add overload for the `inplace` argument.

### DIFF
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -109,7 +109,7 @@ class DataFrame:
     @overload
     def dropna(
         self,
-        inplace: Literal[False] = False,
+        inplace: Literal[False] = ...,
         axis: int = ...,
         how: _str = ...,
         subset: _ColSubsetType = ...,
--- a/pandas-stubs/core/frame.pyi
+++ b/pandas-stubs/core/frame.pyi
@@ -33,6 +33,8 @@ _ListLike = Union[Series, Index, _np.ndarray, Sequence]
 
 _DType = TypeVar("_DType", bound=_np.dtype)
 
+_ColSubsetType = Union[Series, DataFrame, List[_str], _str, _np.ndarray[_np.str_]]
+
 class DataFrame:
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


pandas-stubs/core/frame.pyi
```diff
@@ -104,7 +106,22 @@ class DataFrame:
...
+        subset: _ColSubsetType = ...,
+        inplace: Literal[False] = ...,
+    ) -> DataFrame: ...
+    @overload
+    def dropna(
+        self,
+        axis: int = ...,
+        how: _str = ...,
+        subset: _ColSubsetType = ...,
+        inplace: Literal[True] = ...,
```
olliethomas:
```suggestion
        inplace: Literal[True],
```


Hi, thanks for the PR :)
With the current type signatures mypy has no way of distinguishing between the 2 `dropna` signatures as both have default values on the only argument which changes (`inplace`).
We can disambiguate by making `inplace` only be required when it's not the default value of `False`.

Not sure that makes sense, let me know if it just sounds like a ramble 😅 

Also, if you accept this suggestion, you'll need to re-order the args  so that inplace comes after self.

rpgoldman:
@olliethomas -- Was trying to fix the error you found in my PR, #72, but GitHub won't let me comment on it.  Any chance the protections are wrong?

Anyway, I tried to fix it, but it didn't work out, and I'm not able to replicate the `mypy` test locally: on my laptop I get a host of *other* errors, but not the error I'm seeing on Travis. So I'm kind of limited to pushing changes and hoping.

rpgoldman:
Oh, I see -- this was coming from a GitHub transient flakiness.  AFAICT with your fixes this is good to go.

pandas-stubs/core/frame.pyi
```diff
@@ -104,7 +106,22 @@ class DataFrame:
     def count(self) -> Series: ...
     def drop(self, index: Union[List[_str], Index], axis: _AxisType = ...) -> DataFrame: ...
     def drop_duplicates(self, keep: Union[_str, bool] = ...) -> DataFrame: ...
-    def dropna(self, axis: int = ..., how: _str = ...) -> DataFrame: ...
+    @overload
+    def dropna(
+        self,
+        inplace: Literal[False] = False,
```
olliethomas:
```suggestion
        inplace: Literal[False] = ...,
```

False is already defined in the type signature :)

olliethomas:
Merged! Glad to have you aboard 🚀

rpgoldman:
> Merged! Glad to have you aboard 🚀

Cool!  Excited to have these stubs!

Issue #73, opened 2020-02-26
False positives on `np.empty`


rpgoldman:
I have these two calls:
```
    master_df[DF_VAR_COLUMN] = np.empty(shape=master_df.shape[0], dtype=str)
    master_df[DF_VAR_IDX_COLUMN] = np.empty(shape=master_df.shape[0], dtype=int)
```

With the numpy stubs in place, mypy does not like this, claiming that:
```
eval.py:809: error: Value of type variable "_DType" of "empty" cannot be "str"
eval.py:810: error: Value of type variable "_DType" of "empty" cannot be "int"
```

But these are legitimate values, AFAICT.  I'll try to see about a  PR.

PR #75, opened 2020-02-26
Support str and int as dtypes.

rpgoldman:
Extend the set of dtypes with the str and int literals.

Note -- it would help to add some comments to describe the intended use of the _Dtype types -- it was hard for me to guess if I needed to also extend any of these.

Fix for #73


rpgoldman:
In its current form, this does not work, because I don't know how to specify "the literal that is the class object for the string class."  Mypy seems to treat `Literal[str]` as if I am mistakenly passing an incorrect argument to `Literal`.

Note that I tried this because I was making empty numpy arrays of strings and integers, using `dtype=str` and `dtype=int`, but maybe it would be better to choose some other way to tell numpy what I want, instead of further extending the `_Dtype` type?

rpgoldman:
I tried using `np.str_` and `np.int_` in my code to replace simple `str` and `int`, but mypy with the data science stubs does not seem to like this, either.

thomkeh:
Do you have a code example for what you want to do?

This should already work:

```python
a: np.ndarray[np.int64] = np.array([], dtype=int)
```
This doesn't work, but could be added relatively easily:
```python
a: np.ndarray[np.str_] = np.array([], dtype=str)
```
I'm pretty sure that `Literal` is the wrong choice here, but I'm not sure about your use case.

(It's true that there should be better comments in the stubs.)

rpgoldman:
My use case was `np.empty(shape=master_df.shape[0], dtype=str)`.  See issue #73 

If I replace `int` with `np.int_` and `str` with `np.str_`, that make mypy happy (I  had to add `int_` to the set of types in `numpy_stubs/__init__.py`)

thomkeh:
I see! There is unfortunately no easy solution to this. The only way to do this (as far as I know at least) is to manually add `@overload`s to the `np.empty` function:

```python
@overload
def empty(shape: _ShapeType, dtype: Type[_Int]) -> ndarray[_Int]: ...
@overload
def empty(shape: _ShapeType, dtype: Type[_Float]) -> ndarray[_Float]: ...
@overload
def empty(shape: _ShapeType, dtype: Type[str_]) -> ndarray[str_]: ...
@overload
def empty(shape: _ShapeType, dtype: Type[bool]) -> ndarray[bool_]: ...
@overload
def empty(shape: _ShapeType, dtype: Type[int]) -> ndarray[int64]: ...
@overload
def empty(shape: _ShapeType, dtype: Type[float]) -> ndarray[float64]: ...
@overload
def empty(shape: _ShapeType, dtype: Type[str]) -> ndarray[str_]: ...
```

The reason why this is so complicated is that you cannot tell mypy: "please automatically map `str` to `np.str_`, `int` to `np.int64`, `float` to `np.float64` and `bool` to `np.bool_`". I searched for a nicer way to do this for a long time, but the only thing I found was to do lots of `@overload`s.

rpgoldman:
I am putting in your `empty()` stubs, but I am wondering -- should I make the one with `Type[float]` be the default, since that's what we get if we don't specify?

Also, although I know it should theoretically be a separate PR, I am adding a stub for `np.log10` copied from yours for `np.log2`.

rpgoldman:
`object` is also an ok value for `dtype`.  Should we just treat it as the same as `Any`?

thomkeh:
> I am putting in your empty() stubs, but I am wondering -- should I make the one with Type[float] be the default, since that's what we get if we don't specify?

yes, we can have one overload where dtype is not specified:

```python
def empty(shape: _ShapeType) -> ndarray[float64]: ...
```

> `object` is also an ok value for dtype. Should we just treat it as the same as `Any`?

that's another very good question. We used to have `object` as one of the allowed dtypes, but it cause all sorts of problems. The problem is that `int`, `float` and so on are all subtypes of `object`. This made mypy really confused about what the dtype should be. We can try again to add `object`, but it should be a separate PR.

> Also, although I know it should theoretically be a separate PR, I am adding a stub for `np.log10` copied from yours for `np.log2`.

That's fine 👍

rpgoldman:
> > `object` is also an ok value for dtype. Should we just treat it as the same as `Any`?
> 
> that's another very good question. We used to have `object` as one of the allowed dtypes, but it cause all sorts of problems. The problem is that `int`, `float` and so on are all subtypes of `object`. This made mypy really confused about what the dtype should be. We can try again to add `object`, but it should be a separate PR.

I have been mostly using `object` as a synonym for `np.str_`.  I think if I just don't use it, I'll be happier!

Issue #74, opened 2020-02-26
"at" method is missing from pandas DataFrame stub


rpgoldman:
I'm working on a PR for this.

PR #76, opened 2020-02-26
Add support for Series and DF at methods

rpgoldman:
Created _AtIndexer classes for Series and DataFrame and used them to type the corresponding at() methods.

Partial solution to #74.

This doesn't fully work, because it doesn't handle the possibility that a data frame will contain categorical (string) or integer data, instead of just float.  I don't know how to do this.


pandas-stubs/core/frame.pyi
```diff
@@ -210,3 +210,11 @@ class DataFrame:
...
     def update(self, other: Union[DataFrame, Series]) -> None: ...
     def where(self, cond: Union[Series, DataFrame, _np.ndarray]) -> DataFrame: ...
+    @property
+    def at(self) -> _AtIndexerFrame: ...
+
+# Local Variables:
+# blacken-line-length: 100
+# blacken-allow-py36: t
+# blacken-skip-string-normalization: t
+# End:
```
thomkeh: which editor produced this? it's pretty interesting. we can maybe keep it.

pandas-stubs/core/indexing.pyi
```diff
@@ -83,6 +83,18 @@ class _LocIndexerFrame:
         value: Union[DataFrame, Series, float],
     ) -> None: ...
 
+class _AtIndexerFrame:
+    # get item
+    def __getitem__(self, idx: Tuple[int, _StrLike]) -> float: ...
```
thomkeh: you were concerned about returning just `float` here, right? It's a big problem with DataFrame. I spent a lot of time thinking about it, but couldn't find a perfect solution. The problem is that a DataFrame can have multiple types. Each column can have a different type.

_Ideally_, we would make DataFrame generic with _all_ the column types. Something like:

```python
df: DataFrame[["column_a", int], ["column_b", float]] = pd.DataFrame(
    [[3, 0.3], [4, 0.4]], columns=["column_a", "column_b"]
)
x: int = df["column_a"].iloc[0]
y: float = df["column_b"].iloc[1]
```

But this would make the type stubs *super complicated*, if it is even possible.

So, for now, I think we cannot get around returning a `Union` here:

```suggestion
    def __getitem__(self, idx: Tuple[int, _StrLike]) -> Union[int, float, str]: ...
```

Returning `Union` always sucks because then you don't know what the result is, but I don't see a better solution.

pandas-stubs/core/indexing.pyi
```diff
@@ -83,6 +83,18 @@ class _LocIndexerFrame:
...
+    # set item
+    def __setitem__(self, idx: Tuple[int, _StrLike], value: float) -> None: ...
+
+class _AtIndexerSeries:
```
thomkeh: For the Series object, the situation is easier: it can only have one type. So we'll make this generic in the dtype:
```suggestion
class _AtIndexerSeries(Generic[_DType]):
```

pandas-stubs/core/indexing.pyi
```diff
@@ -83,6 +83,18 @@ class _LocIndexerFrame:
...
+    # get item
+    def __getitem__(self, idx: _StrLike) -> float: ...
```
thomkeh:
```suggestion
    def __getitem__(self, idx: _StrLike) -> _DType: ...
```

use the generic dtype

pandas-stubs/core/series.pyi
```diff
@@ -150,3 +150,11 @@ class Series(Generic[_DType]):
     def unique(self) -> _np.ndarray: ...
     def update(self, other: Series) -> None: ...
     def value_counts(self, normalize: bool = False) -> Series[_DType]: ...
+    @property
+    def at(self) -> _AtIndexerSeries: ...
```
thomkeh:
```suggestion
    def at(self) -> _AtIndexerSeries[_DType]: ...
```

pass on the dtype to the indexer object

rpgoldman:
I applied your changes and I think this is ready for (squash) merge.

rpgoldman:
@thomkeh I think that this is ready for another review, and possibly merging.

pandas-stubs/core/frame.pyi
```diff
@@ -94,13 +97,18 @@ class DataFrame:
...
     @overload
     def apply(self, f: Callable[..., int]) -> Series: ...
     @overload
     def apply(self, f: Callable[..., _ListLike]) -> DataFrame: ...
-    def astype(self, dtype: _str) -> DataFrame: ...
+    def astype(
+        self,
+        dtype: Union[_str, Type[_np.dtype], Dict[_str, Type[_np.dtype]]],
+        copy: bool = True,
+        errors: _ErrorType = 'raise',
```
thomkeh: We follow the recommendation of not putting the default values into the type stubs.

```suggestion
        copy: bool = ...,
        errors: _ErrorType = ...,
```

pandas-stubs/core/frame.pyi
```diff
@@ -21,26 +21,29 @@ import numpy as _np
...
 _AxisType = Literal["columns", "index"]
 
+_ErrorType = Literal['raise', 'ignore']
+
 _ListLike = Union[Series, Index, _np.ndarray, Sequence]
 
 _DType = TypeVar("_DType", bound=_np.dtype)
 
-_ColSubsetType = Union[Series, DataFrame, List[_str], _str, _np.ndarray[_np.str_]]
+_ColSubsetType = Union[Series, 'DataFrame', List[_str], _str, _np.ndarray[_np.str_]]
```
thomkeh: This quoting shouldn't be needed in type stub files. mypy has multiple parsing phases, so it doesn't matter whether `DataFrame` has not been defined at this point. Did you get an error here?

```suggestion
_ColSubsetType = Union[Series, DataFrame, List[_str], _str, _np.ndarray[_np.str_]]
```

numpy-stubs/__init__.pyi
```diff
@@ -465,6 +468,14 @@ def fromiter(iterable: Iterator, dytpe: Type[_DType], count: int = ...) -> ndarr
 def fromstring(
     string: str, dtype: Type[_DType] = ..., count: int = ..., sep: str = ...
 ) -> ndarray: ...
+def histogramdd(
+    a: ndarray,
+    bins: Optional[Union[ndarray, Series, List, int]],
+    range: Optional[List[Tuple[number, number]]] = None,
+    density: bool = False,
+    normed: bool = False,
+    weights: Optional[Union[ndarray, Series, List[number]]] = None,
```
thomkeh: Same thing with the default values
```suggestion
    range: Optional[List[Tuple[number, number]]] = ...,
    density: bool = ...,
    normed: bool = ...,
    weights: Optional[Union[ndarray, Series, List[number]]] = ...,
```

numpy-stubs/__init__.pyi
```diff
@@ -465,6 +468,14 @@ def fromiter(iterable: Iterator, dytpe: Type[_DType], count: int = ...) -> ndarr
 def fromstring(
     string: str, dtype: Type[_DType] = ..., count: int = ..., sep: str = ...
 ) -> ndarray: ...
+def histogramdd(
+    a: ndarray,
+    bins: Optional[Union[ndarray, Series, List, int]],
```
thomkeh: I think we should maybe use [a protocol](https://mypy.readthedocs.io/en/stable/protocols.html) here instead of `Series`. A protocol describes which method a class should implement. For example, `Iterable` describes all classes that have an `__iter__` function. The protocol that we should use here instead of Series, should be [Sequence](https://docs.python.org/3/library/typing.html#typing.Sequence). We would need to check that `Series` really is a `Sequence`. A `Sequence` needs to have these functions: `__len__`, `__iter__`, `__contains__`, `__getitem__`, `__reversed__`, I think. Maybe only `__len__` and `__getitem__` actually. I would need to try it out.

I can understand though if you don't want to wait anymore on this getting merged. So maybe we'll do it in another PR.

numpy-stubs/__init__.pyi
```diff
@@ -42,6 +44,7 @@ class float64(floating): ...
 class float32(float64): ...
 class integer(number, int): ...
 class int64(integer): ...
+class int_(int64): ...
```
thomkeh:
```suggestion
```

doesn't seem to be needed.

pandas-stubs/__init__.pyi
```diff
@@ -8,7 +8,7 @@ from .core.frame import DataFrame as DataFrame
 from .core.series import Series as Series
 from .core.indexes import Index as Index
 
-_AxisType = Literal["columns", "index"]
+_AxisType = Literal["columns", "index", 0, 1]
```
thomkeh: If you change this here, you should probably change the definition of `_AxisType` everywhere else as well...

numpy-stubs/__init__.pyi
rpgoldman: I agree with both your points -- I think we need to use `Sequence` more widely -- I just found that we really need that instead of `List` for the parameter types in `np.array`.  I also think we should push that out to a separate commit: this one's already too big.

pandas-stubs/__init__.pyi
rpgoldman: I will change it in `frame.pyi` and then import it into `__init__.pyi` from `frame`.

numpy-stubs/__init__.pyi
rpgoldman: Added #81  so this can be tracked more easily.

rpgoldman:
Resolved conflicts and force-pushed.

rpgoldman:
@thomkeh merge?

thomkeh:
Sorry, I was very busy with work! Looks all good!

PR #82, opened 2020-03-06
WIP: Make NestedList admit sequences.

rpgoldman:
numpy will actually accept not just lists, but also tuples as array-like.

Discovered this by getting a type error on code that worked correctly.


rpgoldman:
This does not work fully in its current form, because of signatures overlap:

```
/opt/hostedtoolcache/Python/3.7.6/x64/lib/python3.7/site-packages/numpy-stubs/__init__.pyi:417: error: Overloaded function signatures 4 and 8 overlap with incompatible return types
/opt/hostedtoolcache/Python/3.7.6/x64/lib/python3.7/site-packages/numpy-stubs/__init__.pyi:421: error: Overloaded function signatures 6 and 8 overlap with incompatible return types
Found 2 errors in 1 file (checked 3 source files)
```

These two signatures:

* number four
  ```
  @overload
  def array(object: _NestedList[bool]) -> ndarray[bool_]: ...
  ```
  and
* number six
  ```
  @overload
  def array(object: _NestedList[int]) -> ndarray[int64]: ...
  ```

are reported to overlap with number 6:
```
@overload
def array(object: _NestedList[float]) -> ndarray[float64]: ...
```

The second one makes sense to me, because an `int` is a `float`, and I made the variable in the (now misnamed) `_NestedList` definition be covariant.

But I'm confused about the first one, because this seems to suggest that a `bool` is also a `float`, and that seems wrong.  Of course, I don't know if the indices mypy prints are zero-based or one-based, so maybe I am misinterpreting the error message.

The reason I ended up doing this was that I elsewhere had a type defined as `_BoolInt = Literal[0, 1]` and I wanted a list with a `_BoolInt` to be understood as something that could be translated into an `ndarray[int]`.  But to do that required that the type variable used to define `_NestedList` would have to be covariant, which I believe caused this problem.

I will try removing the covariant specification, and just add a cast to the code I'm type-checking.

rpgoldman:
Removing the `covariant` specification leads to *more* errors in the mypy test, not less. So I could use some help figuring out what is wrong here, and how to fix it.

rpgoldman:
@thomkeh This is an attempt to move towards using `Sequence` where the existing code has `List`, which is not as inclusive as numpy is.

But as you can see, it doesn't work in its current form, and I'm not sure why. If someone could review this, I'd appreciate any help.

PR #84, opened 2020-03-17
Fixes some type errors for pandas-stubs

aguillon:
I think I spotted a few mistakes in pandas-stubs, namely:
 * the use of 'axis' in DataFrame.drop is improper when index is specified, because this argument is ignored (set to 1). Therefore, I don't think it should be part of the signature. I also an overloaded version for columns.
 * pd.Indexes don't accept all iterables, for example 'str' is not allowed. I suggest to accept Series, ndarray and List as arguments. In the doc, this types are called "1-D array-like", hence I propose a new type synonym as well. This also requires the following:
 * pd.Series accepts strs (treated as a scalar). I added _str in the corresponding _DType.
 * 'str' was still used as a type instead of '_str' in some places. I did not make an overall check.

I also propose a signature for DataFrame.insert.

Please tell me if anything is wrong.


thomkeh:
I would suggest the following signature for `drop`:

```python
    @overload
    def drop(self, labels: Union[List[_str], Index], axis: _AxisType = ...) -> DataFrame: ...
    @overload
    def drop(self, *, index: Union[List[_str], Index]) -> DataFrame: ...
    @overload
    def drop(self, *, columns: Union[_str, List[_str], Index]) -> DataFrame: ...
```

(the `*` forces named arguments)

this would cover these cases:

```python
# 1. overload
df.drop(["a"])
df.drop(["a"])

aguillon:
@thomkeh that's a clever use of *, thanks! I updated my fork.

aguillon:
Rebased so that aguillon:master is up to date.

PR #88, opened 2020-04-01
Corrected index_col arg of read_csv to be a 5-way Union type.

MylesBartlett:


pandas-stubs/__init__.pyi
```diff
@@ -32,7 +32,7 @@ def read_csv(
     filepath_or_buffer: Union[str, Path],
     sep: str = ...,
     header: Optional[int] = ...,
-    index_col: Optional[int] = ...,
+    index_col: Optional[Union[str, int, Sequence[str], Sequence[int], bool]] = ...,
```
thomkeh, line 35: don't use `Sequence[str]`

thomkeh, line 35: never use `Sequence[str]`

thomkeh:
also, it recommended to use imperative voice in commit messages

pandas-stubs/__init__.pyi
MylesBartlett: Couldn't you have made this a suggestion?

thomkeh, line 35:
```suggestion
    index_col: Optional[Union[str, int, List[str], Tuple[str, ...], Sequence[int], bool]] = ...,
```

MylesBartlett, line 35:
```suggestion
    index_col: Optional[Union[str, int, Union[List[str], Tuple[str, ...]], Sequence[int], bool]] = ...,
```

thomkeh:
you have a nested Union now

PR #89, opened 2020-04-01
Add stubs for MultiIndex attributes and factory functions.

MylesBartlett:
Add stubs for FrozenList.


pandas-stubs/core/indexes/base.pyi
```diff
@@ -64,3 +68,39 @@ class Index(Generic[_T]):
...
+    def names(self) -> FrozenList[str]: ...
+    @property
+    def levels(self) -> FrozenList[Index[_T]]: ...
+    @property
+    def codes(self) -> FrozenList[_np.ndarray[_np.int8]]: ...
+    @property
+    def nlevels(self) -> int: ...
+    @property
+    def levshape(self) -> Tuple[int, ...]: ...
+    def from_arrays(
```
thomkeh, line 83:
```suggestion
    @classmethod
    def from_arrays(
```

thomkeh: the other from_* methods as well

pandas-stubs/core/indexes/base.pyi
```diff
@@ -64,3 +68,39 @@ class Index(Generic[_T]):
...
+        self,
```
thomkeh, line 84:
```suggestion
    @classmethod
    def from_arrays(
```

thomkeh: the other from_* methods as well

pandas-stubs/core/indexes/frozen.pyi
```diff
@@ -0,0 +1,18 @@
+from typing import Generic, Iterator, List, overload, TypeVar, Tuple, Union
+
+__all__ = ["FrozenList"]
```
thomkeh, line 3: you don't really need this, because the things you import from typing are not re-exported (they're not using the `as ...` import) and the variable `_T` starts with an underscore, so that's also not exported.

thomkeh: I would slightly prefer if you removed all these `__all__`, they're not used anywhere else in these type stubs

PR #90, opened 2020-04-01
Make MultiIndex available under pd.MultiIndex

thomkeh:


pandas-stubs/core/__init__.pyi
```diff
@@ -0,0 +1 @@
+from . import indexes as indexes
```
thomkeh, line 1:
```suggestion
```

pandas-stubs/core/__init__.pyi
```diff
@@ -0,0 +1 @@
+
```
thomkeh, line 1:
```suggestion
```

PR #94, opened 2020-04-07
Introduce ArrayLike data type.

MylesBartlett:


thomkeh:
As you can see (from the test that I've added), `np.concatenate` can't actually be called with scalars. I suspected this and thus suggested adding a true `ArrayLike` type and a `ScalarLike` type to cover the rest.

numpy-stubs/__init__.pyi
```diff
@@ -572,9 +578,7 @@ def ceil(a: _FloatObj) -> _FloatObj: ...
 @overload
 def ceil(a: ndarray[_DType]) -> ndarray[_DType]: ...
 def clip(a: ndarray[_DType], a_min: _DType, a_max: _DType) -> ndarray[_DType]: ...
-def concatenate(
-    an: Union[List[ndarray[_DType]], Tuple[ndarray[_DType], ...]], axis: _AxesType = ...
-) -> (ndarray[_DType]): ...
+def concatenate(arrays: Sequence[_ArrayLike[_DType]], axis: _AxesType = ...) -> ndarray[_DType]: ...
```
thomkeh, line 581: scalars are not allowed here

PR #95, opened 2020-04-10
Pyplot stubs

MylesBartlett:


matplotlib-stubs/pyplot.pyi
```diff
@@ -113,4 +115,38 @@ def subplots_adjust(
 ) -> None: ...
 def close(fig: Union[Figure, Literal["all"]]) -> None: ...
 def clf() -> None: ...
+def plot(*args: Any, scalex: bool = ..., scaley: bool = ..., data: Optional[_Data] = ..., **kwargs: Any) -> Line2D:
```
thomkeh, line 118: Please copy the definition of `plot()` from `axes.pyi`. And then add any additional parameters as needed.

matplotlib-stubs/pyplot.pyi
```diff
@@ -113,4 +115,38 @@ def subplots_adjust(
...
+    y2: Optional[Union[list, _np.ndarray]] = ...,
+    where: Optional[List[bool]] = ...,
+    interpolate: bool = ...,
+    step: Optional[Literal["pre", "post", "mid"]] = ...,
+    *,
+    data: Optional[_Data] = ...,
+    **kwargs: Any,
+) -> PolyCollection: ...
+def axhline(y: Optional[float] = ..., xmin: Optional[float] = ..., xmax: Optional[float] = ...) -> Line2d: ...
+def axvline(x: Optional[float] = ..., ymin: Optional[float] = ..., ymax: Optional[float] = ...) -> Line2d: ...
```
thomkeh, line 152: Are you sure these can all be `None`?

matplotlib-stubs/pyplot.pyi
```diff
@@ -113,4 +115,38 @@ def subplots_adjust(
 ) -> None: ...
 def close(fig: Union[Figure, Literal["all"]]) -> None: ...
 def clf() -> None: ...
+def plot(*args: Any, scalex: bool = ..., scaley: bool = ..., data: Optional[_Data] = ..., **kwargs: Any) -> Line2D:
 def show() -> None: ...
+def xlim(*args: Any, **kwargs: Any) -> Tuple[float, float]: ...
```
thomkeh, line 120:
```suggestion
# TODO: figure out more useful definition
def xlim(*args: Any, **kwargs: Any) -> Tuple[float, float]: ...
```

matplotlib-stubs/collections.pyi
```diff
@@ -0,0 +1,11 @@
+from .artist import Artist
+from .cm import ScalarMappable
+
+class Collection(Artist, ScalarMappable):
+    pass
+
+class _CollectionWithSizes(Collection):
```
thomkeh, line 7: I don't think we need to include private classes in our stubs

matplotlib-stubs/pyplot.pyi
```diff
@@ -113,4 +115,38 @@ def subplots_adjust(
...
+def ylabel(
+    ylabel: str,
+    fontdict: Optional[Dict[str, Union[str, int]]] = ...,
+    labelpad: Optional[float] = ...,
+    **kwargs: Any,
+) -> str: ...
+def fill_between(
+    x: Union[list, _np.ndarray],
+    y1: Union[list, _np.ndarray],
+    y2: Optional[Union[list, _np.ndarray]] = ...,
```
thomkeh, line 143: list -> List

thomkeh:
You will also need to add the new files to setup.py. I'm really sorry pull requests for this repository are always so tedious. I wish it was easier to contribute to this library.

matplotlib-stubs/collections.pyi
MylesBartlett: I don't need it as a parent class for PolyCollection?

matplotlib-stubs/pyplot.pyi
MylesBartlett: According to the source code, all the arguments, strangely enough, are Optional.

matplotlib-stubs/collections.pyi
thomkeh: We can just say

```python
class PolyCollection(Collection):
     pass
```
right? skipping the middle man.

matplotlib-stubs/pyplot.pyi
thomkeh: 👍

matplotlib-stubs/collections.pyi
MylesBartlett: While I doubt few, if any, people will make use of them, what about the methods defined in `_CollectionWithSizes`?

matplotlib-stubs/pyplot.pyi
olliethomas: I think it's "optional" in the same sense as the numpy docs say args with default values are "optional" (based on the below which is from [here](https://matplotlib.org/3.1.1/_modules/matplotlib/pyplot.html#axvline))
`def axvline(x=0, ymin=0, ymax=1, **kwargs)`

matplotlib-stubs/collections.pyi
thomkeh: You can leave it as it is. But in general I don't think we should spend our time replicating the minutia of the class hierarchies in matplotlib.

matplotlib-stubs/pyplot.pyi
MylesBartlett: Ah I see, yes rather than the source code I meant the docs.

MylesBartlett: Trying to disentangle the various definitions of array-like and optional is a real pain.

matplotlib-stubs/pyplot.pyi
```diff
@@ -1,17 +1,23 @@
...
+from .axes import SubplotBase, _LegendLocation
+from .collections import PolyCollection
 from .image import AxesImage
+from .legend import Legend
 from .text import Text
 
-_Float = TypeVar("_Float", _np.float32, _np.float64)
-_Data = Union[float, _np.ndarray[_Float], Sequence[float]]
+_T = TypeVar("_T")
+_Numeric = TypeVar("_Numeric", int, float)
```
thomkeh, line 17: https://stackoverflow.com/a/50928627

matplotlib-stubs/pyplot.pyi
```diff
@@ -104,13 +113,88 @@ def figure(
...
+    ymin: _Numeric = ...,
+    ymax: _Numeric = ...,
+) -> Tuple[_Numeric, _Numeric]: ...
+def xticks(
+    ticks: Optional[_NumericArray] = ...,
+    labels: Optional[_StrArray] = ...,
+    *,
+    fontdict: Optional[Dict[str, Union[str, int]]] = ...,
+    minor: bool = ...,
+    **kwargs: Any,
```
thomkeh, line 162: In general I don't like `**kwargs` in type stubs. they make the whole affair kind of pointless...

matplotlib-stubs/pyplot.pyi
```diff
@@ -1,17 +1,23 @@
...
+from .collections import PolyCollection
 from .image import AxesImage
+from .legend import Legend
 from .text import Text
 
-_Float = TypeVar("_Float", _np.float32, _np.float64)
-_Data = Union[float, _np.ndarray[_Float], Sequence[float]]
+_T = TypeVar("_T")
+_Numeric = TypeVar("_Numeric", int, float)
+_Data = Union[_Numeric, _np.ndarray[_Numeric], Sequence[_Numeric]]
```
thomkeh, line 18: `ndarray` may not be parameterized with Python scalars

thomkeh: Wait, why doesn't mypy complain about this?

numpy-stubs/ma.pyi
```diff
@@ -1,7 +1,8 @@
 from numpy import ndarray, _DType, bool_
 from typing import Generic
 
-class MaskedArray(ndarray[_DType]): ...
+class MaskedArray(ndarray[_DType]):
+    pass
```
thomkeh, line 5: Wait, why is *this* class with `pass`?

PR #98, opened 2020-04-16
minor updates

olliethomas:


numpy-stubs/__init__.pyi
```diff
@@ -773,10 +773,10 @@ class matrix:
...
 # Specific values
 #
-inf: float
-nan: float
-NaN: float
-NAN: float
+inf: float32
+nan: float32
+NaN: float32
+NAN: float32
```
thomkeh, line 779: this doesn't seem true?

```python
type(np.inf) == float
```

pandas-stubs/core/frame.pyi
```diff
@@ -184,7 +198,7 @@ class DataFrame:
     def rename(self, mapper: Callable, axis: _AxisType = ...) -> DataFrame: ...
     @overload
     def rename(self, columns: Dict[_str, _str]) -> DataFrame: ...
-    def replace(self, a: float, b: float) -> DataFrame: ...
+    def replace(self, a: _DTypeEx, b: _DType, regex: bool = ...) -> DataFrame: ...
```
thomkeh, line 201: Why use a TypeVar here? Just do
```suggestion
    def replace(self, a: Union[_np.dtype, str], b: Union[_np.dtype, str], regex: bool = ...) -> DataFrame: ...
```

TypeVars are only needed if the type of the arguments influence the return type

numpy-stubs/__init__.pyi
olliethomas: yeah....I was trying to get around a problem by introducing a new problem/error. Resolved

pandas-stubs/core/frame.pyi
olliethomas: good point

PR #99, opened 2020-04-20
Define previously missing Figure.savefig params.

MylesBartlett:
Duplicate Figure.savefig to define plot.savefig.


matplotlib-stubs/pyplot.pyi
```diff
@@ -245,9 +246,63 @@ class Figure:
...
+            "svgz",
+            "tif",
+            "tiff",
+        ] = ...,
+        transparent: bool = ...,
+        bbox_inches: Optional[Union[Literal["tight"], Bbox]] = ...,
+        pad_inches: Optional[int] = ...,
         bbox_extra_artists: Sequence[Artist] = ...,
-        bbox_inches: Optional[Literal["tight"]] = ...,
+        metadata: Union[Dict[str, Optional[str]], Dict[ByteString, ByteString]] = ...,
```
thomkeh, line 303: is it really `ByteString`? Can it not be a string? Are you maybe looking for [AnyStr](https://docs.python.org/3/library/typing.html#typing.AnyStr)?

MylesBartlett: Really the metadata keys should be a literal but I cba at the moment define a literal for each of the backends.  The documentation for the metadata arg for the pdf backend reads as follows: "Metadata in the PNG file as key-value pairs of bytes or latin-1 encodable strings". For the other two backends string key-value pairs are expected, so far as I can tell.

PR #100, opened 2020-04-22
added some functions that I've come across in my travels.

olliethomas:


numpy-stubs/__init__.pyi
```diff
@@ -58,6 +58,7 @@ _DType2 = TypeVar("_DType2", bool_, float32, float64, int8, int16, int32, int64,
 _DTypeObj = TypeVar("_DTypeObj", bound=Union[dtype, int, float])
 _ShapeType = Union[int, Tuple[int, ...], List[int]]
 _AxesType = Union[int, Tuple[int, ...], List[int]]
+_InterpolationType = Literal['linear', 'lower', 'higher', 'midpoint', 'nearest']
```
olliethomas, line 61: I added this as it's own type rather than have the literal in the function signature, but it's only used in the `percentile` and `quantile` functions (and I haven't even typed quantile yet), so I completely understand if you would prefer this in the function signature.

olliethomas:
```
@overload
def atleast_2d(ary: Union[ndarray, Sequence[ndarray]]) -> ndarray: ...
@overload
def atleast_2d(*ary: ndarray) -> List[ndarray]: ...
```
it was always gonna be too good to be true
`error: Overloaded function signatures 1 and 2 overlap with incompatible return types`

olliethomas:
https://numpy.org/doc/stable/reference/generated/numpy.atleast_2d.html

thomkeh:
This should work:

```python
@overload
def atleast_2d(ary: Union[ndarray, Sequence[ndarray]]) -> ndarray: ...
@overload
def atleast_2d(*ary: Union[ndarray, Sequence[ndarray]]) -> List[ndarray]: ...
```

it has to be the same type in both cases.

numpy-stubs/__init__.pyi
thomkeh: No, it's fine. I like defining this elsewhere. Keeps the function signature short.

numpy-stubs/__init__.pyi
```diff
@@ -641,6 +660,7 @@ def matmul(a: ndarray[_DType], b: ndarray[_DType]) -> ndarray[_DType]: ...
 def max(a: ndarray[_DType], axis: None = ...) -> _DType: ...
 @overload
 def max(a: ndarray[_DType], axis: _AxesType, keepdims: bool = ...) -> ndarray[_DType]: ...
+def maximum(a: ndarray, b: ndarray) -> ndarray: ...
```
thomkeh, line 663:
```suggestion
def maximum(a: ndarray[_DType], b: ndarray[_DType]) -> ndarray[_DType]: ...
```

...unless you're worried about comparing ndarrays of different types?

numpy-stubs/__init__.pyi
```diff
@@ -662,6 +682,7 @@ def median(
 def min(a: ndarray[_DType], axis: None = ...) -> _DType: ...
 @overload
 def min(a: ndarray[_DType], axis: _AxesType, keepdims: bool = ...) -> ndarray[_DType]: ...
+def minimum(a: ndarray, b: ndarray) -> ndarray: ...
```
thomkeh, line 685:
```suggestion
def minimum(a: ndarray[_DType], b: ndarray[_DType]) -> ndarray[_DType]: ...
```

same here

numpy-stubs/__init__.pyi
```diff
@@ -773,10 +804,12 @@ class matrix:
 #
 # Specific values
 #
+e: float
```
thomkeh, line 807: huh, I didn't know this existed. But it makes sense.

numpy-stubs/__init__.pyi
```diff
@@ -349,6 +354,7 @@ class Array(Protocol[_DType]):
     def __array__(self) -> Union[ndarray[_DType], Sequence[Sequence[_DType]]]: ...
 
 _ArrayLike = Union[Array[_DType], Sequence[_DType]]
+_Coercable = Union[_ArrayLike, _DTypeObj]
```
olliethomas, line 357: Added as `atleast_2d` can have inputs of `_ArrayLike`, or values that are not `_ArrayLike` , but that can be coerced into an array

tests/numpy_test.py
```diff
@@ -118,3 +118,22 @@ def test_concatenate() -> None:
...
+    a = np.atleast_2d(1, [1, 2], [[1, 2]], 1)
+    b = [
+        np.array([[1]]),
+        np.array([[1, 2]]),
+        np.array([[1, 2]]),
+        np.array([[1]]),
+    ]
+    assert len(a) == len(b)
+    for ex_a, ex_b in zip(a, b):
+        assert (ex_a == ex_a).all()
```
olliethomas, line 139: I just copied the example from the numpy docs of how to use this function. I can get rid of the value assertions if you like? Numpy should be testing itself

thomkeh: Yeah we don't really care about the values. We only care about dtypes and whether something is an ndarray or a scalar.

numpy-stubs/__init__.pyi
olliethomas: btw, I didn't think you could coerce a str, but apparently you can 🤷‍♂️

numpy-stubs/__init__.pyi
```diff
@@ -575,6 +590,10 @@ def asarray(a: ndarray, dtype: Type[float]) -> ndarray[float64]: ...
 @overload
 def asarray(a: ndarray, dtype: Type[str]) -> ndarray[str_]: ...
 @overload
+def atleast_2d(ary: _Coercable) -> ndarray: ...
+@overload
+def atleast_2d(ar: _Coercable, *ary: _Coercable) -> List[ndarray]: ...
```
thomkeh, line 595: I'm guessing it didn't work with just `*ary`?

tests/numpy_test.py
```diff
@@ -118,3 +118,22 @@ def test_concatenate() -> None:
     d3: np.ndarray[np.int32] = np.concatenate((d, d), axis=0)
     scalar: np.float32 = np.float32(3.0)
     assert isinstance(scalar, np.float32)
+
+
+def test_at_least_2d() -> None:
+    assert np.atleast_2d(3.0) == np.array([[3.0]])
```
thomkeh, line 124:
```suggestion
    arr: np.ndarray[np.float64] = np.atleast_2d(3.0)
    assert isinstance(arr, np.ndarray)
    assert_dtype(arr, np.float64)
```

tests/numpy_test.py
```diff
@@ -118,3 +118,22 @@ def test_concatenate() -> None:
...
+
+    x = np.arange(3.0)
+    assert (np.atleast_2d(x) == np.array([[0.0, 1.0, 2.0]])).all()
+
```
thomkeh, line 128:
```suggestion
```

tests/numpy_test.py
thomkeh, line 139:
```suggestion
    a = np.atleast_2d(1, [1, 2], [[1, 2]], 1)
    assert isinstance(a, list)
```

numpy-stubs/__init__.pyi
```diff
@@ -672,7 +696,17 @@ def nan_to_num(
 ) -> ndarray[_DType]: ...
 def nonzero(a: ndarray) -> Tuple[ndarray[int64], ...]: ...
 def outer(a: ndarray[_DType], b: ndarray[_DType]) -> ndarray[_DType]: ...
-def percentile(a: ndarray[_DType], q: float) -> ndarray[_DType]: ...
+@overload
+def percentile(
+    a: ndarray[_DType], q: float, interpolation: _InterpolationType = ..., axis: _AxesType = ...
```
thomkeh, line 701:
```suggestion
    a: ndarray[_DType], q: _FloatObj, interpolation: _InterpolationType = ..., axis: _AxesType = ...
```

numpy-stubs/__init__.pyi
olliethomas: No, I think Mypy must treat `*arr` as 1 or more `arr` because it couldn’t distinguish between the 2 signatures. I say this is Mypy behaviour as I would’ve thought `*arr` would mean 0 or more, but that might be me getting it wrong.

thomkeh: I think the problem is just that the  return types are incompatible. Like, I'm pretty sure this

```python
@overload
def f(ary: int) -> Tuple[int]: ...
@overload
def f(*ary: int) -> Tuple[int, ...]: ...
```
would have worked because `Tuple[int]` is a subclass of `Tuple[int, ...]`.

The reasoning by mypy is this: if the first overload is a special case of the second, then the return type of the first should also be a special case of the return type of the second. But `ndarray` is not a special case of `List[ndarray]`, so it didn't work.

The reasons behind that behavior of mypy are explained here: https://mypy.readthedocs.io/en/stable/more_types.html#type-checking-the-variants

olliethomas:
ooh, if you go to files changed you can select a number of suggestions and apply them all in 1 commit. Nice 😎

tests/numpy_test.py
```diff
@@ -118,3 +118,12 @@ def test_concatenate() -> None:
...
     scalar: np.float32 = np.float32(3.0)
     assert isinstance(scalar, np.float32)
+
+
+def test_at_least_2d() -> None:
+    arr: np.ndarray[np.float64] = np.atleast_2d(3.0)
+    assert isinstance(arr, np.ndarray)
+    assert_dtype(arr, np.float64)
+
+    a = np.atleast_2d(1, [1, 2], [[1, 2]], 1)
```
thomkeh, line 128: Very minor improvement. Not worth delaying merging for.

```suggestion
    a: List[np.ndarray[np.int64]] = np.atleast_2d(1, [1, 2], [[1, 2]], 1)
```

PR #105, opened 2020-05-19
Acs 2

olliethomas:


olliethomas:
this definitely will be full of mistakes, but pushing as requested

numpy-stubs/__init__.pyi
```diff
@@ -280,6 +280,7 @@ class ndarray(Generic[_DType]):
         self,
         key: Union[
             None,
+            range,
```
thomkeh, line 283: doesn't seem to be used anywhere
```suggestion
```

numpy-stubs/random.pyi
```diff
@@ -43,6 +54,9 @@ def exponential(scale: _FloatObj, size: Sequence[_IntObj]) -> ndarray[float64]:
 def exponential(scale: Sequence[_FloatObj], size: Sequence[_IntObj]) -> ndarray[float64]: ...
 def geometric(p: float, size: _IntObj) -> ndarray[float64]: ...
 def get_state() -> Tuple[str, ndarray[_Int], _IntObj, _IntObj, _FloatObj]: ...
+def multinomial(
+    n: int, pvals: _ArrayLike[float] = ..., size: Union[int, Tuple[int, ...]] = ...
```
thomkeh, line 58:
```suggestion
    n: int, pvals: _ArrayLike[_Float] = ..., size: Union[int, Tuple[int, ...]] = ...
```

this should at least get rid of the error. not sure if it's correct though

PR #115, opened 2020-06-19
Pin mypy to an older version

thomkeh:


olliethomas:
Cool, but how come?

thomkeh:
@olliethomas see the error from this comment: https://github.com/predictive-analytics-lab/data-science-types/pull/116#issuecomment-647977761

PR #116, opened 2020-06-21
Shelelena/pandas improvements

Shelelena:
Improvements in pandas DataFrame, DataFrameGroupBy and SeriesGroupBy

- specify DataFrame.groupby
- add DataFrameGroupBy.aggregate
- adjust data type in DataFrame.\_\_init\_\_
- add \_\_getattr\_\_ to get columns in DataFrame and DataFrameGroupBy
- correct return type of DataFrameGroupBy.\_\_getitem\_\_
- add some missing statistical methods to DataFrameGroupBy and SeriesGroupBy


pandas-stubs/core/groupby/generic.pyi
```diff
@@ -1,21 +1,51 @@
...
     def max(self) -> Series: ...
     def mean(self) -> Series: ...
     def median(self) -> Series: ...
     def min(self) -> Series: ...
+    def nunique(self, dropna: bool = ...) -> Series: ...
+    def quantile(self, q: float = ..., interpolation: str = ...) -> Series: ...
     def std(self, ddof: int = ...) -> Series: ...
+    def sum(self) -> Series: ...
+    def tail(self, n: int = ...) -> Series: ...
+    def unique(self) -> Series[_np.ndarray]: ...
```
thomkeh, line 50: This line throws an error in the mypy check. Does it really return a Series of numpy arrays?

thomkeh:
In order to run tests, you should be able to do
```
./check_all.sh
```
(after you've installed `black`, `mypy`, `flake8` and `flake8-pyi`)

pandas-stubs/core/groupby/generic.pyi
Shelelena: It does, here's the [documentation](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.core.groupby.SeriesGroupBy.unique.html). And you could try this code:
```
sepal_length = iris.groupby('species').sepal_length
type(sepal_length.unique())
type(sepal_length.unique()[0])
```
I could change that to simple Series return type, should I?

Shelelena: I would rather add _np.ndarray to _DType

Shelelena:
Thanks for this

> In order to run tests, you should be able to do
> 
> ```
> ./check_all.sh
> ```
> 
> (after you've installed `black`, `mypy`, `flake8` and `flake8-pyi`)

I've added changes made by black and removed np.ndarray specification. (Though it's possible to get it back later along with the Series specification)

I still get an error locally when using `mypy tests`:
```
.../env/lib/python3.8/site-packages/mypy/typeshed/stdlib/3/typing.pyi:675: error: Return type becomes "Union[bool, Any]" due to an unfollowed import
Found 1 error in 1 file (checked 3 source files)
```
What other dependencies do I need to avoid this?

pandas-stubs/core/groupby/generic.pyi
Shelelena: I removed np.ndarray specification for now

thomkeh:
> I still get an error locally when using mypy tests:

That's an issue with the latest mypy version. You can ignore it or use mypy version 0.770. (The problem is that there is a mistake in the type stubs for the standard library. These type stubs are bundled together with mypy.)

pandas-stubs/core/groupby/generic.pyi
thomkeh: Ok, I tried your code and is indeed a Series of ndarray. So, yeah, this should be added to `_DType` in `series.pyi`. I'm leaving it up to you if you want to do that. Otherwise I'll merge it as it is right now.

pandas-stubs/core/frame.pyi
```diff
@@ -164,9 +167,33 @@ class DataFrame:
...
+    def groupby(
+        self,
+        by: Union[Series[_str], Dict[_str, _str], Callable],
+        axis: _AxisType = ...,
+        level: Union[int, _str] = ...,
+        sort: bool = ...,
+        group_keys: bool = ...,
+        squeeze: bool = ...,
+        observed: bool = ...,
+    ) -> DataFrameGroupBy: ...
```
thomkeh, line 196: wait, these two overloads now return the same thing. what happened to the overload that returns `SeriesGroupBy`?

pandas-stubs/core/frame.pyi
```diff
@@ -37,10 +37,12 @@ _DType = TypeVar("_DType", bound=_np.dtype)
 
 _ColSubsetType = Union[Series, DataFrame, List[_str], _str, _np.ndarray[_np.str_]]
 
+_FunctionLike = Union[_str, Callable]
```
thomkeh, line 40: This doesn't seem to be used anywhere

thomkeh: oh it's used in a different file

pandas-stubs/core/frame.pyi
Shelelena: Yeah that's right, the `groupby()` never returned a `SeriesGroupBy` as long as I tested it, and so says the [documentation](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.groupby.html). The only way to get a `SeriesGroupBy` was to extract a column from a `DataFrameGroupBy`. If you succeed in getting a `SeriesGroupBy` from the `groupby()` method, could you please provide the code?

The two versions of `groupby` you are seeing now differ by the axis.
Single index (which may be multiindex) and list of indices work with the default `axis=0`, and the `as_index` argument is allowed, like this:
```
iris.groupby('species', as_index=False)
```
The second version work with mappings, that may be `dict`, `Series` or `function`, it requires `axis=1` and does not allow `as_index=False`. The example (dropping the species column, which is not numeric, may be required for aggregating):
```
iris.groupby({
    'sepal_length': 'length',
    'petal_length': 'length',
}, axis=1)
```
Both versions return `DataFrameGroupBy`.

pandas-stubs/core/groupby/generic.pyi
Shelelena: I've finally returned the `np.ndarray` specification :)
I also added `list` to the `_DTypes`: it is returned by `Series.str.split()` method, for example.

pandas-stubs/core/frame.pyi
thomkeh: I see! Thanks for investigating this!

thomkeh:
@Shelelena I created a new release with your changes: 0.2.15. Thanks again for the PR! And thanks for fixing the existing issues as well! I really appreciate it.

Shelelena:
@thomkeh I'm glad to be of service :)
Your library is very useful for me. I suppose I'll contribute a bit more :)

Shelelena:
I've just read that your philosophy is not to recreate the documentation (sorry I didn't read it earlier). I think I overdid it with the groupby method. Don't you mind if I make it less verbose next time?

thomkeh:
It's definitely completely fine if you don't include all methods and all possible arguments (only the most important ones). The task of writing type stubs for pandas is so huge that we have to accept imperfect type stubs, otherwise we will never make any progress at all. But, completely recreating the original code is also never wrong! If you're willing to do it, you're very welcome to ;)

However, there are cases where we have to deliberately deviate from the actual code. For example, a numpy ndarray can contain `object`s. But if we allow this, then all type checks become useless, because _any_ value can be interpreted as `object`. So we made the decision to disallow `object`s in an ndarray.

But, yeah, it's definitely okay if you make it less verbose next time :)

PR #121, opened 2020-06-26
Multi python

olliethomas:


.github/workflows/mypy_check.yml
```diff
@@ -3,24 +3,42 @@ name: Mypy
 on:
   pull_request:
     branches:
-    - master
+    - multi-python
```
olliethomas, line 6: eurgh - missed this

olliethomas, line 6:
```suggestion
    - master
```

setup.py
```diff
@@ -56,7 +56,9 @@
     ],
     python_requires=">=3.6",
     # use `pip install data-science-types[dev]` to install development packages
-    extras_require={"dev": ["black", "mypy", "flake8", "flake8-pyi"]},
+    extras_require={
```
olliethomas, line 59: btw, did a new environment to run the 36 tests locally. Couldn't run without installing these extra packages for the tests, so added them

check_all.sh
```diff
@@ -5,6 +5,6 @@ echo "Begin check..." \
 && python gen_pyi.py \
 && black -l 100 -t py36 --pyi --fast matplotlib-stubs/pyplot.pyi.in \
 && flake8 *-stubs \
-&& python -m pytest -vv tests/ \
-&& mypy tests \
+&& python -m pytest -vv tests/py37+ \
```
olliethomas, line 8: btw, this is where I was thinking of using the `python -V` to have a conditional about which tests to run/check based on the python version.

olliethomas:
Looking at the checks - this needs the branch protection rules updating

thomkeh:
I think you should remove the tests for python 3.7+ and use only the ones for python 3.6. They will still work with 3.7+ and I don't like maintaining two almost identical sets of tests.

Yeah and then we'll need to change the protection rules, but let's only do that once #122 has been merged.

.github/workflows/test_correctness.yml
```diff
@@ -3,23 +3,40 @@ name: Pytest
 on:
   pull_request:
     branches:
-    - master
+    - multi-python
```
thomkeh, line 6:
```suggestion
    - master
```

thomkeh:
closed in favor of #123

PR #122, opened 2020-06-27
Shelelena/pandas.DataFrame fixes

Shelelena:
* add `DataFrame.tail()` and DataFrame.idxmin()
* add default argument value to `DataFrame.head()` and `DataFrame.idxmax()`
* change `DataFrame.mean()` and `DataFrame.std()` return type to `Series`
* refactor `DataFrame.sum()` to get rid of `overload`
* update `DataFrame.rename()` so than it can rename indices and use both dicts and callables as argument
* fix `Type[_np.dtype]` not working as `dtype` argument in `DataFrame` methods, add `_TypeLike`
* fix `DataFrame.insert()` `value` argument, remove `_DType`, which is no longer used


pandas-stubs/core/frame.pyi
```diff
@@ -35,19 +35,19 @@ _ErrorType = Literal["raise", "ignore"]
 
 _ListLike = Union[Series, Index, _np.ndarray, Sequence]
 
-_DType = TypeVar("_DType", bound=_np.dtype)
-
 _ColSubsetType = Union[Series, DataFrame, List[_str], _str, _np.ndarray[_np.str_]]
 
 _FunctionLike = Union[_str, Callable]
 
+_TypeLike = Union[_str, _np.dtype, Type[float], Type[_str]]
```
thomkeh, line 42: This should be
```suggestion
_TypeLike = Union[_str, Type[_np.dtype], Type[float], Type[_str]]
```
right?

pandas-stubs/core/frame.pyi
```diff
@@ -196,21 +196,18 @@ class DataFrame:
...
-        allow_duplicates: bool = ...,
+        self, loc: int, column: _str, value: _ListLike, allow_duplicates: bool = ...,
     ) -> None: ...
     def isin(self, values: Union[Iterable, Series, DataFrame, Dict]) -> DataFrame: ...
     def isna(self) -> DataFrame: ...
     def isnull(self) -> DataFrame: ...
     def iterrows(self) -> Generator[Series, Series, Series]: ...
     def max(self) -> Series: ...
-    def mean(self) -> DataFrame: ...
+    def mean(self) -> Series: ...
```
thomkeh, line 210: Thanks for catching this embarrassing mistake!

pandas-stubs/core/frame.pyi
```diff
@@ -223,9 +220,11 @@ class DataFrame:
     def query(self, expr: _str) -> DataFrame: ...
     def reindex(self, index: Index) -> DataFrame: ...
     @overload
-    def rename(self, mapper: Callable, axis: _AxisType = ...) -> DataFrame: ...
+    def rename(self, mapper: Union[Dict, Callable], axis: _AxisType = ...) -> DataFrame: ...
```
thomkeh, line 223: Can we say anything about the key and value type of the `Dict`? Maybe
```python
Dict[Union[str, int], Union[str, int]]
```

(This is actually related to a larger problem: what type can column identifier have? Is it only string and int?)

pandas-stubs/core/frame.pyi
```diff
@@ -223,9 +220,11 @@ class DataFrame:
...
     @overload
-    def rename(self, columns: Dict[_str, _str]) -> DataFrame: ...
+    def rename(self, columns: Union[Dict, Callable]) -> DataFrame: ...
+    @overload
+    def rename(self, index: Union[Dict, Callable]) -> DataFrame: ...
```
thomkeh, line 227: Can we [force named arguments](https://python-3-for-scientists.readthedocs.io/en/latest/python3_advanced.html) here?

```suggestion
    def rename(self, *, index: Union[Dict, Callable]) -> DataFrame: ...
```

Otherwise this is a bit confusing.

pandas-stubs/core/frame.pyi
Shelelena: Nope, check the code:
```iris.astype(np.dtype('str'))```
vs
```iris.astype(np.dtype)```

`Type[_np.dtype]` expects the second one, but it raises `TypeError`

File: pandas-stubs/core/frame.pyi (hunk 3 above)
Comment (Shelelena) [reply pandas-stubs/core/frame.pyi:223]: We could restrict it to str and int as the most common cases.
I suppose the column identifier can have any hashable type (for ex. tuple is allowed, while list is not), so I we could also restrict it to
```
Dict[Hashable, Hashable]
```

pandas-stubs/core/frame.pyi
Shelelena: You're right, added this for both index and columns argument

pandas-stubs/core/frame.pyi
thomkeh: `Hashable` sounds great!

pandas-stubs/core/frame.pyi
thomkeh: While looking into this, I realized our numpy type hierarchy is wrong. So, let's merge it as it is, and then I'll fix the numpy dtype problem.

pandas-stubs/core/frame.pyi
Shelelena: done :)

thomkeh:
👍

Shelelena:
Thanks :)

PR #123, opened 2020-06-29
Run tests for multiple python versions

thomkeh:
I decided on 3.6 as it is the oldest we support and 3.8 as it is the newest we support.

Supersedes #121

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,17 @@ You can install these dev dependencies via
 pip install -e .[dev]
 ```
 
+This will also install numpy, pandas and matplotlib to be able to run the tests.
+
+### Pre-pull request checks
+
+We also include a script that runs (most of) the CI checks that will be run when a PR is opened.
+To test these out locally, use the `check_all.sh` script.
+
+```bash
+./check_all.sh
+```
+
 ### Checking compliance with Mypy
 The settings for Mypy are specified in the `mypy.ini` file in the repository.
 Just running
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ This will also install numpy, pandas and matplotlib to be able to run the tests.
 
 ### Pre-pull request checks
 
-We also include a script that runs (most of) the CI checks that will be run when a PR is opened.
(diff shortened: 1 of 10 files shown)


thomkeh:
This still leaves us with this failure:

```
    def test_indexing_with_df() -> None:
>       b: pd.DataFrame = df[s.to_frame().isin([df])]

tests/pandas_test.py:73: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/hostedtoolcache/Python/3.6.10/x64/lib/python3.6/site-packages/pandas/core/frame.py:8442: in isin
    algorithms.isin(self.values.ravel(), values).reshape(self.shape),
/opt/hostedtoolcache/Python/3.6.10/x64/lib/python3.6/site-packages/pandas/core/algorithms.py:453: in isin
    return f(comps, values)
pandas/_libs/hashtable_func_helper.pxi:437: in pandas._libs.hashtable.ismember_object
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    def __hash__(self):
        raise TypeError(
>           f"{repr(type(self).__name__)} objects are mutable, "
            f"thus they cannot be hashed"
        )
E       TypeError: 'DataFrame' objects are mutable, thus they cannot be hashed
```
which no-one understands.

README.md
```diff
@@ -90,6 +90,17 @@ You can install these dev dependencies via
 pip install -e .[dev]
 ```
 
+This will also install numpy, pandas and matplotlib to be able to run the tests.
+
+### Pre-pull request checks
+
+We also include a script that runs (most of) the CI checks that will be run when a PR is opened.
```
thomkeh, line 97:
```suggestion
We also include a script that runs the CI checks that will be run when a PR is opened.
```

PR #125, opened 2020-07-04
Add `nanstd` variant of `std`

AnnyZhao:
Reference: https://numpy.org/doc/stable/reference/generated/numpy.nanstd.html

I copied the definitions from `std` in this file.


thomkeh:
Thanks :)

PR #126, opened 2020-07-20
Add `empty` property for pandas.DataFrame

herberthamaral:
The title is self-explanatory. We're missing this property there.


thomkeh:
Thanks!

PR #127, opened 2020-07-25
A few methods on DataFrame, Series,  Index, added or refined

davidatbu:
Hi all,

Thanks for the great effort towards typing numpy and pandas! It's much appreciated.

I modified some lines on a fork because they were relevant to code I was writing recently. I thought I might as well submit a PR.


pandas-stubs/core/frame.pyi
```diff
@@ -207,7 +211,12 @@ class DataFrame:
     def isin(self, values: Union[Iterable, Series, DataFrame, Dict]) -> DataFrame: ...
     def isna(self) -> DataFrame: ...
     def isnull(self) -> DataFrame: ...
-    def iterrows(self) -> Generator[Series, Series, Series]: ...
+    def iterrows(self) -> Iterator[Tuple[_Label, Series]]: ...
+    def itertuples(
+        self, index: bool = ..., name: Optional[_str] = ...
+    ) -> Iterator[
+        Tuple[Any, ...]
+    ]: ...  # Don't know how to express a named tuple with a variable name/num of slots
```
thomkeh, line 219: What happens if you return `NamedTuple` (from `typing`)? Does mypy complain that it doesn't have any entries?

pandas-stubs/__init__.pyi
```diff
@@ -39,7 +39,10 @@ def read_csv(
     squeeze: bool = ...,
     prefix: Optional[str] = ...,
     mangle_dupe_cols: bool = ...,
-    dtype: Optional[Union[Type, str, Dict[str, Union[Type, str]]]] = ...,
+    # Having Dict[str, Union[Type, str]] instead of
+    # Union[Dict[str, Type], Dict[str, str]] breaks the "Liskov substitution
+    # principle" and will result in mypy complaining when passing in a dicts.
+    dtype: Optional[Union[Type, str, Dict[str, Type], Dict[str, str]]] = ...,
```
thomkeh, line 45: Would using `Mapping` instead of `Dict` help? (`Mapping` is covariant in the value.) I think some people legitimately pass things like `{"a": str, "b": "float32"}` to this function.

davidatbu: Yes. `Mapping` helped.

pandas-stubs/core/frame.pyi
davidatbu: When I tried `NamedTuple`, Mypy infers `Any` as the type of the expression `next(df.itertuples(name='some name'))`.

davidatbu:
Pinging @thomkeh in case notifications regarding my changes in response to your comments fell through the cracks

thomkeh:
It did indeed fall through, sorry!

Issue #130, opened 2020-08-02
types missing (e.g. uint32)


Maarten-vd-Sande:
I am new to type-hinting, but I thought I will give it a go :)

I noticed that quite some types are missing:
https://numpy.org/doc/stable/user/basics.types.html
https://github.com/predictive-analytics-lab/data-science-types/blob/master/numpy-stubs/__init__.pyi#L43

Is there a reason for these values missing (except that it is a lot of work to migrate all at once)? I might be able to free up some time to add these in a PR if you are interested.

olliethomas:
Hi Maarten, thanks for the question! 

Numpy is a massive library, which is great, but it posed a problem to us - if we want to have type hints for numpy, how should we do this? As you've already said, doing it all at once is a lot of work. Instead, we decided to just add type hints for the parts of numpy that we were using in our projects, and over time add new, or expand upon existing, type hints as and when this becomes useful. 

It's been really nice that other people have found this and decided to contribute too, adding type hints for the parts of numpy, pandas and matplotlib that they've been using. Because of these contributions this project has made steps towards typing these massive libraries, but there's a loooong way to go. 

I guess this is all a long way of saying, yes, please do open a PR - we always welcome contributions, and are happy to help you either get started, or help solve some of the trickier problems.

Welcome aboard!

Oliver

PR #131, opened 2020-08-03
add support for unsigned integers

Maarten-vd-Sande:
As mentioned briefly in #130, I added support for unsigned integers. When adding unsigned + unsigned -> unsigned, but unsigned + signed -> signed.

Now the still missing datatypes are:
np.intp, np.uintp, np.complex64, np.complex128

But I am not entirely sure how to add them, and they are quite niche (even more so than unsigned ;) )


Maarten-vd-Sande:
Ah, let me add some tests!

thomkeh:
type-hinting tests are a big problem. I'm not really happy with how we're doing them, but I haven't found a better solution.

That said, something like this would be nice (in the numpy test file):

```python
def test_unsigned() -> None:
    x: np.ndarray[np.uint8] = np.array([3, 4], dtype=np.uint8)
    y: np.ndarray[np.uint8] = x + x
```

Maarten-vd-Sande:
@thomkeh I added the tests. Thanks for the help :)

thomkeh:
Great, thanks!

PR #144, opened 2020-08-15
Allow DataFrame.sort_values by a single column name

melentye:
according to the documentation, by= can be a str or a list of str


thomkeh:
Thanks!

PR #151, opened 2020-08-26
add np.diff

### LINKED ISSUE #150: missing np.diff
https://numpy.org/doc/stable/reference/generated/numpy.diff.html

clouds56:
fix #150


numpy-stubs/__init__.pyi
```diff
@@ -665,6 +665,13 @@ def corrcoef(
 ) -> ndarray[float64]: ...
 def cov(m: ndarray[_DType], rowvar: Optional[bool]) -> ndarray[float64]: ...
 def diag(a: ndarray[_DType]) -> ndarray[_DType]: ...
+def diff(
+    a: ndarray[_DType],
+    n: _Int = ...,
+    axis: _AxesType = ...,
+    prepend: Optional[ndarray[_DType]] = ...,
```
thomkeh, line 672:
```suggestion
    prepend: ndarray[_DType] = ...,
```

This is a bit confusing. In the numpy documentation, "optional" just means it's a parameter with a default value. But `typing.Optional` means that the value can be `None`. But `prepend` may not be `None`; I tried it out:

```
In [1]: import numpy as np

In [2]: np.diff(np.array([2, 3, 4]), n=1, prepend=None)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-37ada9a52273> in <module>
----> 1 np.diff(np.array([2, 3, 4]), n=1, prepend=None)

<__array_function__ internals> in diff(*args, **kwargs)

~/anaconda/envs/pytorch/lib/python3.8/site-packages/numpy/lib/function_base.py in diff(a, n, axis, prepend, append)
   1267     op = not_equal if a.dtype == np.bool_ else subtract
   1268     for _ in range(n):
-> 1269         a = op(a[slice1], a[slice2])
   1270
   1271     return a

TypeError: unsupported operand type(s) for -: 'int' and 'NoneType'
```

numpy-stubs/__init__.pyi
```diff
@@ -665,6 +665,13 @@ def corrcoef(
...
+    append: Optional[ndarray[_DType]] = ...,
```
thomkeh, line 673:
```suggestion
    append: ndarray[_DType] = ...,
```

numpy-stubs/__init__.pyi
```diff
@@ -665,6 +665,13 @@ def corrcoef(
 ) -> ndarray[float64]: ...
 def cov(m: ndarray[_DType], rowvar: Optional[bool]) -> ndarray[float64]: ...
 def diag(a: ndarray[_DType]) -> ndarray[_DType]: ...
+def diff(
+    a: ndarray[_DType],
+    n: _Int = ...,
```
thomkeh, line 670: This has to be `_IntObj` because `_Int` is reserved for the dtype of ndarrays.
```suggestion
    n: _IntObj = ...,
```

thomkeh (approved):
Thanks! There were just a few tricky bits that I commented on.

Issue #152, opened 2020-08-26
[pandas] `_AtIndexerFrame` should support indexing by [int, int]


ZeeD:
if you build a Dataframe with `header=None` the axes are `[RangeIndex(start=0, stop=4, step=1), RangeIndex(start=0, stop=9, step=1)]`, so you can't access elements with `df.at[123, 'xyz']`, but you need to use `df.at[123,456]`.

according to https://github.com/predictive-analytics-lab/data-science-types/blob/7dab8238df9e93d00be6d683d8efabbdf95fc958/pandas-stubs/core/indexing.pyi#L88, however, `_AtIndexerFrame` now only allow the second index to be a "_StrLike".

Ideally this should be a runtime check, but as I thing this is not possible with mypy, there should not be a false positive

ZeeD:
@thomkeh Thanks for the merge

thomkeh:
Thanks for sending a PR!

PR #154, opened 2020-08-26
let int indices for column labels for the df.at property

ZeeD:
fix to issue #152


thomkeh:
Ah, the old issue that column names can also be integers. I think there was a previous PR where we determined that the correct type for columns should actually be `Hashable`. So, I would suggest the following:

- rename `_StrLike` to `_ColType` in the whole file
- define `_ColType` like this: `Union[Hashable, _np.str_]` (you will have to import `Hashable` from `typing`)

That should solve your problem.

```
In [1]: from typing import Hashable

In [2]: isinstance(6, Hashable)
Out[2]: True

In [3]: isinstance(6.4, Hashable)
Out[3]: True

In [4]: isinstance([], Hashable)
Out[4]: False
```

ZeeD:
isn't `_np.str_` also `Hashable`?

```python
>>> from typing import Hashable
>>> import numpy as _np
>>> isinstance(_np.str_(), Hashable)
True
```

can I just use `Hashable` and drop `_StrLike` / `_ColType`?

thomkeh:
Ok, sure.

ZeeD:
uhm... I'm having trouble understanding the black check... putting aside the fact that the fails are on files/rows I didn't touched, I can't reproduce the same error on my machine... what is the exact command launched?

Surely, I can apply the diff, but is it something that should go in this PR?

thomkeh:
The reason was the `black` pushed a new release which changed how it works. And GitHub CI always just downloads the latest version. Maybe we should pin one version of black.

PR #160, opened 2020-08-29
Adds type information for pandas.DataFrame.transpose

### LINKED ISSUE #159: Add type information for pandas.DataFrame.transpose.
There is currently no type information for `pandas.DataFrame.transpose`.

lmcnichols:
Fixes #159


thomkeh:
Hi, thanks for the PR! I'm not sure `Optional` is the right type here though. In Python's `typing` library, `Optional` means that the value may be `None`. But I don't think these values can be `None`, can they?

The fact that the values have _default values_ is already encoded by the `= ...`.

pandas-stubs/core/frame.pyi
```diff
@@ -143,6 +143,7 @@ class DataFrame:
     @overload
     def drop(self, *, columns: Union[_str, List[_str], Index]) -> DataFrame: ...
     def drop_duplicates(self, keep: Union[_str, bool] = ...) -> DataFrame: ...
+    def transpose(self, *args: Optional[int], copy: Optional[bool] = ...) -> DataFrame: ...
```
thomkeh, line 146:
```suggestion
    def transpose(self, *args: int, copy: bool = ...) -> DataFrame: ...
```

PR #168, opened 2020-09-18
Flesh out pandas and numpy a bit more

TheCleric:
This is the result of testing data-science-types against another project I contribute to: https://github.com/jldbc/pybaseball

* I added some common .gitignores for venv and vscode
* I found a few Pandas tweaks to support functions and parameters that we are using.
    * Tweak DataFrame.apply, DataFrame.drop, DataFrame.merge, DataFrame.rank, DataFrame.reindex, DataFrame.replace
    * Add DataFrame.assign, DataFrame.filter
    * Tweak Series.rank
    * Add pandas.isnull
    * Tweak DataFrame.loc
* A few changes to numpy as well
    * Allow tuples -> numpy.array
    * Tweak numpy.setdiff1d
    * Add numpy.cos, numpy.deg2rad, numpy.sin, numpy.cos

Everything was done using the latest Pandas docs for reference to data types:

https://pandas.pydata.org/pandas-docs/stable/reference/

I also did my best to add tests to support the changes as well


TheCleric:
Looks like I'm having a few issues in the github action that I am not having locally:

* Pandas can accept a re.Pattern, which I have a version_info check around, but the action is still flagging it.
* MyPy seems to think the __getitem__ overload I added to _iLocIndexerFrame is superfluous, but my tests fail without it (and I only added it because MyPy was complaining about my code's usage of passing a Tuple[slice, List[str]] to a .loc of a DataFrame, which is valid

Please advise.

pandas-stubs/core/frame.pyi
```diff
@@ -36,7 +40,7 @@ _ErrorType = Literal["raise", "ignore"]
 
 _ListLike = Union[Series, Index, _np.ndarray, Sequence]
 
-_ColSubsetType = Union[Series, DataFrame, List[_str], _str, _np.ndarray[_np.str_]]
+_ColSubsetType = Union[Series, "DataFrame", List[_str], _str, _np.ndarray[_np.str_]]
```
thomkeh, line 43: Quoting DataFrame shouldn't be needed. Mypy does multiple passes of the stub files and can manage undefined references.

pandas-stubs/core/frame.pyi
```diff
@@ -123,10 +127,10 @@ class DataFrame:
     def append(
         self, s: Union[DataFrame, Dict[_str, Any]], ignore_index: bool = ..., sort: bool = ...
     ) -> DataFrame: ...
-    @overload
-    def apply(self, f: Callable[..., int]) -> Series: ...
-    @overload
-    def apply(self, f: Callable[..., _ListLike], axis: _AxisType = ...) -> DataFrame: ...
```
thomkeh, line 129: So, can this return a series?

pandas-stubs/core/groupby/generic.pyi
```diff
@@ -9,11 +10,34 @@ _str = str  # needed because Series has a property called "str"...
...
+    def sum(self) -> Series: ...
+    def tail(self, n: int = ...) -> Series: ...
+    def unique(self) -> Series[_np.ndarray]: ...
+    def var(self, ddof: int = ...) -> Series: ...
+
 class DataFrameGroupBy(GroupBy):
     @overload
     def __getitem__(self, item: _str) -> SeriesGroupBy: ...
     @overload
-    def __getitem__(self, item: List[_str]) -> DataFrameGroupBy: ...
```
thomkeh, line 16: Here as well, no quoting.

pandas-stubs/core/frame.pyi
TheCleric: Yes:

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.apply.html

TheCleric:
> Looks like I'm having a few issues in the github action that I am not having locally:
> 
> * Pandas can accept a re.Pattern in python 3.7+, which I have a version_info check around, but the action is still flagging it.
> * MyPy seems to think the **getitem** overload I added to _iLocIndexerFrame is superfluous, but my tests fail without it (and I only added it because MyPy was complaining about my code's usage of passing a Tuple[slice, List[str]] to a .loc of a DataFrame, which is valid
> 
> Please advise.

Can I get some feedback on fixing these issues so this PR can become mergable?

thomkeh:
Sorry for the delay! So, let's try to figure out the issues.

> Pandas can accept a re.Pattern in python 3.7+, which I have a version_info check around, but the action is still flagging it.

Is it 3.7+ or 3.6+? In the code you wrote `if sys.version_info >= (3, 6):`. If it was 3.6+, then we can drop the check, because we only support Python >= 3.6 anyway.

TheCleric:
> Sorry for the delay! So, let's try to figure out the issues.
> 
> > Pandas can accept a re.Pattern in python 3.7+, which I have a version_info check around, but the action is still flagging it.
> 
> Is it 3.7+ or 3.6+? In the code you wrote `if sys.version_info >= (3, 6):`. If it was 3.6+, then we can drop the check, because we only support Python >= 3.6 anyway.

It is 3.7, let me fix that version check.

TheCleric:
Looks like I was able to fix the first part with using Pattern from typing instead of re.

Now the only thing that remains is the overload issue.

TheCleric:
🎉  🥳

thomkeh:
Oh, well done!

PR #180, opened 2020-10-11
add rand and np.nan*

clouds56:


numpy-stubs/__init__.pyi
```diff
@@ -887,6 +885,14 @@ def where(condition: _ConditionType, x: float, y: float) -> ndarray[float64]: ..
 @overload
 def where(condition: _ConditionType) -> Tuple[ndarray[int64], ...]: ...
 
+#
+# nan series methods
+#
+nancumsum = cumsum
+nanmean = mean
+nanstd = std
+nansum = sum
```
thomkeh, line 894: This is pretty clever. I'm trying to think of situations where this breaks something, but they're all quite unlikely. So, let's go with this for now.

thomkeh:
Thanks!

PR #184, opened 2020-10-23
Test with python 3.9 instead of 3.8

thomkeh:
This way we're testing on the oldest still supported (3.6) and the newest.


thomkeh:
This is blocked by one of matplotlib's dependencies (kiwi solver) not supporting Python 3.9 yet.

PR #186, opened 2020-10-24
Fix signature of numpy.load and numpy.save

nicoddemus:
According to the docs, both [save](https://numpy.org/doc/stable/reference/generated/numpy.save.html) and [load](https://numpy.org/doc/stable/reference/generated/numpy.load.html) accept either `str`, `PathLike`, or a file-like object.


thomkeh:
Thanks!

PR #187, opened 2020-10-29
Add average function of numpy

courentin:
This PR add the [average](https://numpy.org/doc/stable/reference/generated/numpy.average.html) function of numpy.
I'm not 100% sure the signature is right though.


thomkeh:
I think you can just copy the signature of `np.mean`? Returning a `Union` is not so nice, because then the type checker always has to deal with all the types in the union.

PR #188, opened 2020-11-05
Add signature for numpy.set_printoptions()

pmav99:


numpy-stubs/__init__.pyi
```diff
@@ -932,6 +932,24 @@ class finfo(Generic[_Float]):
...
+    edgeitems: Optional[Any] = None,
+    linewidth: Optional[Any] = None,
+    suppress: Optional[Any] = None,
+    nanstr: Optional[Any] = None,
+    infstr: Optional[Any] = None,
+    formatter: Optional[Any] = None,
+    sign: Optional[Any] = None,
+    floatmode: Optional[Any] = None,
+    *,
+    legacy: Optional[Any] = None,
```
thomkeh, line 950: If you're using `Any`, there is no need to use `Optional` as well (`Optional` means that the value may be `None`). Also, we don't put the default values into the type stubs, we just use `...`.
```suggestion
    precision: Any = ...,
    threshold: Any = ...,
    edgeitems: Any = ...,
    linewidth: Any = ...,
    suppress: Any = ...,
    nanstr: Any = ...,
    infstr: Any = ...,
    formatter: Any = ...,
    sign: Any = ...,
    floatmode: Any = ...,
    *,
    legacy: Any = ...,
```

pmav99: thank you

PR #189, opened 2020-11-05
added random number Generator. 

olliethomas:
Was a bit lazy with the function definitions. Most of them, under some situations can return a scalar. I've just got them returning an ndarray


thomkeh:
not sure why we didn't merge this. can you resolve the conflicts?

olliethomas:
The types are a bit lazily done.  But, they are perhaps better than nothing

thomkeh:
I'm sure they're great.

PR #191, opened 2020-11-28
Add signature for np.interp

nicoddemus:
As per docs:

https://numpy.org/doc/stable/reference/generated/numpy.interp.html


thomkeh:
Thanks!

PR #192, opened 2020-11-28
Add Pandas to_parquet signature

jmargeta:
Adding signature for DataFrame.to_parquet(...)
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_parquet.html

This method requires either fastparquet or pyarrow to be installed.
They are not supported on 3.9, yet.

Shall I remove the test completely?

See also:
pyarrow: https://github.com/apache/arrow/pull/8386
numba (which fastparquet depends on): https://github.com/numba/numba/issues/6345


thomkeh:
Hm, yeah, remove the test. Or, though this is maybe a bit unusual, could you comment it out and put a comment above it, which says to uncomment it when Python 3.9 support improves? It's also possible to mark tests to be skipped for certain python versions, but this might be overly complicated.

jmargeta:
Thanks for your feedback, @thomkeh, that sounds good.
The PR should now contain only the signature.

thomkeh:
Perfect, thanks!

PR #196, opened 2020-12-02
Added numpy.frombuffer to stub

jtc42:
PR adds support for [numpy.frombuffer](https://numpy.org/doc/stable/reference/generated/numpy.frombuffer.html).

AFAIK there's no nice standard "buffer_like" type yet so right now this uses `Union[bytes, bytearray, memoryview]` which I think covers most cases. Happy to change this in response to feedback though.


thomkeh:
Thanks! Yeah that type looks reasonable.

PR #197, opened 2020-12-03
Implement nanmin and nanmax

Hvlot:
If I understand correctly, I can just add nanmin and nanmax under the nan series methods to introduce these. 

Numpy documentation: 
https://numpy.org/doc/stable/reference/generated/numpy.nanmin.html
https://numpy.org/doc/stable/reference/generated/numpy.nanmax.html


thomkeh:
Yup, that looks right.

PR #198, opened 2020-12-04
Add pd.read_parquet

eganjs:
I have updated the commented out test added by #192 

I did not include the `use_nullable_dtypes` parameter as it was added in Pandas 1.2.0 and data-science-types resolves Pandas to 1.1.4 (on my machine) using a clean Python 3.8.1 venv created by pyenv


thomkeh:
Looks good! Thanks!

PR #203, opened 2020-12-09
DataFrame.sort_values can take a list of bools

sinemetu1:
[See latest documentation here.](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.sort_values.html?highlight=sort_values#pandas.Series.sort_values)


thomkeh:
You linked to the wrong documentation, which confused me at first. But [looking at the correct documentation](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.sort_values.html#pandas.DataFrame.sort_values), it seems you're right! Thanks!

PR #206, opened 2020-12-17
Add Series.sort_index() signature

krassowski:
- [x] Adds `Series.sort_index` based on [stable version documentation](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.sort_index.html#pandas-series-sort-index)
- [x] Fixes wrong order of arguments in `Series.sort_values()` (`ascending` should go before `inplace`)
- [x] Adds missing arguments to `Series.sort_values()`


thomkeh:
Thanks for this!

> Fixes wrong order of arguments in Series.sort_values() (ascending should go before inplace)

Hm, the issue is that if you do an overload based on `inplace=True` and `inplace=False`, then you cannot give both a default value.

That is, consider this:

```python
@overload
def f(inplace: Literal[True] = ...) -> A: ...
@overload
def f(inplace: Literal[False] = ...) -> B: ...
```

and now you call it like this:
```python
x = f()
```
what type does `x` have? We can't tell. So that's why we need to do it like this:
```python
@overload
def f(inplace: Literal[True]) -> A: ...
@overload
def f(inplace: Literal[False] = ...) -> B: ...
```
Then we now that `f()` returns `B`. So, you see, one of the two `inplace` needs to be without default argument. And arguments without default argument need to come _before_ arguments with default argument. That's why `inplace` was in the "wrong" place.

I'm actually very surprised that you managed to make the tests pass with your `inplace` definitions.

krassowski:
The tests pass because the return type for `inplace=False` is set to optional (thus has an overlap with None for `inplace=True`), which is a pattern for overriding inplace that is already used by `fillna` - is that stub also wrong?

krassowski:
Is lack of certainty over the result in the default case (not in place, returning Series, but mypy can only tell is either Series or None) worse than incorrect hints when using positional arguments? As far as I understand the return type will be inferred correctly for non -default in place operation providing useful hint should someone try to use the result (None) as if it was a Series.

So unless I missed something this is not ideal but not wrong either - just a trade-off.

krassowski:
I can obviously revert the `inplace` change if you wish, using positionals for sorting series is not beautiful either way because of to the dummy axis argument

krassowski:
Seems to be working fine now (removing return type overlap, swapping order):

Screenshot from 2020-12-18 01-03-53

Remaining issue; disabling the type check error with a comment does not seem to work great for mypy.

krassowski:
So I think I go it working, what do you think now @thomkeh? I basically accepted that it has to go with `# type: ignore` and as mentioned [here](https://stackoverflow.com/a/53530815/6646912) it is fine because typeshed stubs use this approach to make accurate stubs for built-in libraries, for example see [`encode()` stubs here](https://github.com/python/typeshed/blob/e3352c654d58ffba5951e14aa8972c4c6f17212b/stdlib/2and3/codecs.pyi#L72-L77). Orders matters when matching overloads so the one returning a Series has to go first.

thomkeh:
I made some changes before I merged. I'd really like to avoid `# type: ignore`. I think what I did should cover your use case. Let me know if it doesn't.

krassowski:
Interesting. Seems a bit hacky (I would say more hacky than leaving `# type: ignore`) but it seems to work fine. Thank you!

PR #212, opened 2020-12-22
Added pd.Series.astype(object) overload

danpol:
Minimal example usecase:
```
>>> import pandas as pd
>>> series = pd.Series([1, 2])
>>> series = series.astype(object)
>>> series.iloc[0] = None
>>> print(type(series.values[1]))
<class 'int'>
```
Without `series.astype(object)`, data type would be `<class 'numpy.float64'>`


thomkeh:
Thanks!

PR #213, opened 2020-12-22
Add `isfinite`, `isinf`, `isscalar`, `diagonal`, `allclose`

### LINKED ISSUE #209: Numpy has no method 'isfinite'
Script used:
```python
import numpy as np

arr: np.ndarray[np.float32] = np.array([0, 1, np.inf], dtype=np.float32)

print(np.isfinite(arr))
```

tadeu:
and improve `isnan` support in NumPy stubs. Also add support for scalar + array.

fixes #209


numpy-stubs/__init__.pyi
```diff
@@ -390,6 +390,8 @@ class ndarray(Generic[_DType]):
     def __radd__(self, value: ndarray[_DType]) -> ndarray[_DType]: ...
     @overload
     def __radd__(self, value: _DType) -> ndarray[_DType]: ...
+    @overload
+    def __radd__(self, value: float) -> ndarray[_DType]: ...
```
thomkeh, line 394: I think the return type should be float64?
```suggestion
    def __radd__(self, value: float) -> ndarray[float64]: ...
```

numpy-stubs/__init__.pyi
```diff
@@ -993,6 +991,92 @@ def set_printoptions(
...
+        casting: Any = ...,
+        order: Any = ...,
+        dtype: Any = ...,
+        subok: bool = ...,
+        signature: Union[str, Tuple[str]] = ...,
+        # In reality this should be a length of list 3 containing an
+        # int, an int, and a callable, but there's no way to express
+        # that.
+        extobj: List[Union[int, Callable]] = ...,
+    ) -> Any: ...
```
thomkeh, line 1034: I don't love this Any return type. Is it possible to do an overload like we had before?

tests/numpy_test.py
```diff
@@ -259,3 +259,52 @@ def test_interp() -> None:
...
+
+
+def test_newaxis() -> None:
+    x = np.array([1.0, 2.0])
+    assert x[np.newaxis, :].shape == (1, 2)
+
+
+def test_sum_scalar_before() -> None:
+    x = 273.15 + np.array([-0.1e2, -0.77e1])
+    assert isinstance(x, np.ndarray)
```
thomkeh, line 310: Can we assert the type of the ndarray?

thomkeh (commented):
Thanks! I like the approach with the `ufunc`; we used it for np.mean or np.std before and I always thought that this should be expanded. Also, great tests!

See some comments below.

numpy-stubs/__init__.pyi
tadeu: It´s for the `float + array` case, so the return is also an array

numpy-stubs/__init__.pyi
tadeu: I took this from [NumPy stubs](https://github.com/numpy/numpy/blob/ab585e6f8b047a9ca54578eb93e241917227e985/numpy/__init__.pyi#L2123), I guess the return type from ufuncs is too broad, check this issue: https://github.com/numpy/numpy/pull/17805

tests/numpy_test.py
tadeu: Sure

numpy-stubs/__init__.pyi
tadeu: Hmm, I think that not necessarily... if `_DType` is `np.int64`, the return type will indeed be `np.float64`, but this would not be true for complex types, for example.

Issue #215, opened 2021-01-08
test_frame_iloc fails on Pandas 1.2


EdwardJRoss:
tests/pandas_test.py line 92 fails on Pandas 1.2

Extracting the relevant code

```
import pandas as pd
df: pd.DataFrame = pd.DataFrame(
    [[1.0, 2.0], [4.0, 5.0], [7.0, 8.0]],
    index=["cobra", "viper", "sidewinder"],
    columns=["max_speed", "shield"],
)
s: "pd.Series[float]" = df["shield"].copy()
df.iloc[0] = s
```

Results in

```
ValueError: could not broadcast input array from shape (3) into shape (2)
```

This runs fine on Pandas 1.1.5

PR #217, opened 2021-01-08
Add stubs for to_pickle and read_pickle

### LINKED ISSUE #216: Fails type checking on dataframe.to_pickle
Running mypy with data-science-types on the following

```
import pandas as pd
df = pd.DataFrame({'a': [1]})
df.to_pickle('output.pkl')
```

Produces an error:
```
error: "Series[Any]" not callable
```

I would expect it to pass using [DataFrame.to_pickle](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_pickle.html)

EdwardJRoss:
Resolves #216


thomkeh:
So it seems the tests for python 3.9 are downloading a newer version of pandas (1.2) which has the problem mentioned in #215. Is it easy to fix that?

thomkeh:
I'm actually a bit surprised that pandas already dropped python 3.6 support.

EdwardJRoss:
I guess a workaround would be to use the previous version in the tests by updating [.github/workflows/test_correctness.yml](https://github.com/predictive-analytics-lab/data-science-types/blob/master/.github/workflows/test_correctness.yml#L25) to install `pandas<1.2`

I'm not really clear enough on `iloc` behaviour and the test intention on how to fix the test for Pandas 1.2.

thomkeh:
Okay, then let's do this for now

> updating .github/workflows/test_correctness.yml to install pandas<1.2

You can just edit the file.

thomkeh:
Great, thanks!

PR #221, opened 2021-01-27
Add stub for np.cumprod and np.nancumprod

danhper:
Hi!
`cumprod` was only defined on `ndarray` so I added the function versions: `np.cumprod` and `np.nancumprod`


thomkeh:
Thanks!

PR #223, opened 2021-01-29
Add project URLs

klieret:
Probably want to have links to this repo from the [pypi page](https://pypi.org/project/data-science-types/)


thomkeh:
Thanks!

